PR #198, opened 2021-07-08
Local key store with Postgres database

darioAnongba:
Local key store with Postgres database:
- Implementation of the local key store using an underlying secret store
- Local key store using a PG DB (not implemented)
- PostgreSQL client added

Missing:
- Concreate implementation of postgres DB
- Concrete implementation of postgres client
- Unit tests
- Acceptance tests


.env.sample
```diff
@@ -1,6 +1,6 @@
 LOG_LEVEL=DEBUG
 LOG_FORMAT=text
-HTTP_PORT=8080
+HTTP_PORT=8080HealthKeyManagerURL
```
dolanor, line 3: Is it a typo?

src/stores/infra/postgres/client/dialer.go
```diff
@@ -0,0 +1,140 @@
...
+	"fmt"
+	"io"
+	"net"
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/consensys/quorum-key-manager/pkg/tcp"
+	"github.com/consensys/quorum-key-manager/pkg/tls"
+)
+
+type SSLDialer struct {
```
dolanor, line 16: With even SSL 3.0 deprecated, is it good to have an implem for this?

.env.sample
darioAnongba: removed

src/stores/infra/postgres/client/dialer.go
darioAnongba: it seems the TLSDIaler uses the SSLDIaler:
```
&TLSDialer{
		Dialer: &tls.Dialer{
			Dialer: &SSLDialer{
				Dialer: &net.Dialer{
					Timeout:   cfg.DialTimeout,
					KeepAlive: cfg.KeepAliveInterval,
				},
			},
			TLSConfig: tlsConfig,
		},
		verifyCAOnly: verifyCAOnly,
	}
```

dolanor: Oh, ok, I got it I think.
It actually just needs a wrapper around a `tcp.Dialer`, so maybe rename the SSLDialer to `pgTLSDialer`?
Also, we don't need it outside this package, right? So I guess private visibility would be better?

darioAnongba: comments applied

Makefile
```diff
@@ -27,16 +28,25 @@ hashicorp-down:
 
 networks:
 	@docker network create --driver=bridge hashicorp || true
+	@docker network create --driver=bridge postgres || true
```
ggarri, line 31: Do we need a postgres network? Ideally we should have a qkm network where we attach mandatory services like postgres to. Only in case of optional deps is where we do the otherway around

src/stores/infra/postgres/postgres.go
```diff
@@ -0,0 +1,6 @@
+package postgres
+
+//go:generate mockgen -source=postgres.go -destination=mocks/postgres.go -package=mocks
```
ggarri, line 3: WHat is the propose of mocking this empty interface?

darioAnongba: the interface is not going to stay empty... This is just a skeleton

Makefile
darioAnongba: modified

darioAnongba:
removed zap2, it was an error with automatic refactoring

ggarri (approved):

PR #201, opened 2021-07-09
Implement local key store using PG DB

darioAnongba:
Implement local key store using PG DB:
- Local key store implementation
- Keys data agent implementation
- Infra postgres client implementation
- Acceptance tests using a real Postgres instance


src/stores/store/database/postgres/keys.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+
+	return key, nil
+}
+
+func (d *Keys) GetDeleted(_ context.Context, id string) (*entities.Key, error) {
+	key := &entities.Key{ID: id}
+
+	err := d.db.SelectDeletedPK(key)
```
ggarri, line 43: Every call to DB should depend on ctx otherwise if the routine making the call went death we make DB compute work for not need. Same it would be for any call to external dependencies, context should define the lifetime of the call

src/stores/store/keys/local/local.go
```diff
@@ -0,0 +1,370 @@
...
+		ID:        id,
+		PublicKey: pubKey,
+		Algo:      alg,
+		Metadata:  secret.Metadata,
+		Tags:      attr.Tags,
+	}
+
+	err = s.keysDA.Add(ctx, key)
+	if err != nil {
+		return nil, err
```
ggarri, line 288: if we fail to add it into DB we might have inconsistent state where the key is added into the SecretStore but not to the DB. I would rather implement a rollback strategy here

src/stores/store/keys/local/local.go
```diff
@@ -0,0 +1,370 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = s.keysDA.Restore(ctx, key)
+	if err != nil {
+		return err
+	}
+
+	err = s.keysDA.Add(ctx, key)
```
ggarri, line 163: Why are you adding the key again? As far as I see you already restore it in the store above

src/stores/store/database/postgres/keys.go
darioAnongba: are you sure about that? Because AKV and AWS do not accept context if I'm not mistaken. Also go-pg examples do not use the context.

darioAnongba: added

src/stores/store/entities/metadata.go
```diff
@@ -8,6 +8,6 @@ type Metadata struct {
 	ExpireAt    time.Time
 	CreatedAt   time.Time
 	UpdatedAt   time.Time
-	DeletedAt   time.Time
+	DeletedAt   time.Time `pg:",soft_delete"`
```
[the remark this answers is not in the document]
darioAnongba: model added

src/stores/store/keys/local/local.go
darioAnongba: added db tx

ggarri (approved):

Issue #116, opened 2021-05-31
Add OIDC Authenticator


nmvalera:
#### Summary

Implement OpenID connect Authenticator

#### Acceptance criteria

- [x] Implement OIDC Authenticator under `auth/authenticator/oidc/authenticator.go`
- [x] Create config `auth/authenticator/oidc/config.go` and add it to `auth/manager/config.go`
- [x] Create `viper` configuration flags for `oidc` config and generate config to be pass to app
- [x] Support both Certificate file of the identity server AND Identity server URL to fetch the certificate
- [x] Implement a tool to generate jwts for testing

PR #202, opened 2021-07-09
116 oidc

ggarri:
## PR description
Added support for JWT 

## Fixed Issue(s)
- #116

## Changelog


cmd/flags/auth.go
```diff
@@ -1,10 +1,183 @@
...
+	
+	return &auth.Config{OICD: oicdCfg}, nil
+}
+
+func fileCertificate(vipr *viper.Viper) (string, error) {
+	caFile := vipr.GetString(authOICDCACertFileViperKey)
+	_, err := os.Stat(caFile)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return "", fmt.Errorf("failed to read CA file. %s", err.Error())
```
darioAnongba, line 140: should we use errors from the `errors` package?

cmd/util.go
```diff
@@ -0,0 +1,99 @@
...
+	generateJWTCmd.Flags().StringVar(&username, "username", "", "username added in claims")
+	generateJWTCmd.Flags().StringArrayVar(&groups, "groups", []string{}, "groups added in claims")
+	generateJWTCmd.Flags().DurationVar(&expiration, "expiration", time.Hour, "Token expiration time")
+
+	utilCmd.AddCommand(generateJWTCmd)
+
+	return utilCmd
+}
+
+func runGenerateJWT(_ *cobra.Command, _ []string) error {
```
darioAnongba, line 53: are you using this for tests? not sure we need a JWT generator as part of `cmd`

pkg/auth/auth0.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp := new(JWKsResponse)
+	if err = json2.UnmarshalBody(r.Body, resp); err != nil {
+		return nil, err
+	}
+
+	keyPairs := []certificate.KeyPair{}
```
darioAnongba, line 43: I think that we should be using `https://github.com/square/go-jose` to decode the response and to extract the jwt claims. Not sure why we are defining these types ourselves :thinking: 

We can use ` &jose.JSONWebKeySet{}` and ` &jwt.StandardClaims{}`

darioAnongba, line 43: I believe we can use `https://github.com/square/go-jose` for types and handling of certificates. Check `&jose.JSONWebKeySet{}`.

src/auth/authenticator/oicd/jwt.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, errors.UnauthorizedError(err.Error())
+	}
+	if !token.Valid {
+		return nil, errors.UnauthorizedError("invalid access token")
+	}
+
+	return token.Claims.(*Claims), nil
+}
+
+func (checker *JWTChecker) key(token *jwt.Token) (interface{}, error) {
```
darioAnongba, line 52: You can do something like this:
```
_, ok := token.Method.(*jwt.SigningMethodRSA)
		if !ok {
			errMessage := "invalid access token signing method"
			log(errMessage)
			return nil, errors.New(errMessage)
		}
```

src/auth/authenticator/oicd/testutils/jwtGenerator.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, err
+	}
+
+	return &JWTGenerator{
+		privateKey: cert.PrivateKey.(*rsa.PrivateKey),
+		claims:     claims,
+	}, nil
+}
+
+func (j *JWTGenerator) GenerateAccessToken(username string, groups []string, ttl time.Duration) (tokenValue string, err error) {
```
darioAnongba, line 31: Do we need this in the codebase?

src/auth/authenticator/oicd/authenticator.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return auth, nil
+}
+
+func (a Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// Extract Access Token from context
+	token, ok := extractToken("Bearer ", req.Header.Get("Authorization"))
+	if !ok {
+		return nil, nil
```
darioAnongba, line 45: Does this mean we don't fail when no Bearer token is passed? Shouldn't we return an error and ignore it on the parent?

pkg/auth/auth0.go
```diff
@@ -0,0 +1,51 @@
...
+
+	json2 "github.com/consensys/quorum-key-manager/pkg/json"
+	"github.com/consensys/quorum-key-manager/pkg/tls/certificate"
+)
+
+const (
+	Auth0IssuerServerDomain = "auth0.com"
+)
+
+type JWKsResponse struct {
```
darioAnongba, line 15: I believe we can use `https://github.com/square/go-jose` for types and handling of certificates. Check `&jose.JSONWebKeySet{}`.

src/auth/authenticator/oicd/authenticator.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	auth := &Authenticator{
+		jwtCheckers: []*JWTChecker{},
+	}
+
+	for _, cert := range cfg.Certificates {
+		keyPair, err := certificate.X509KeyPair([]byte(cert), nil)
+		if err != nil {
+			return nil, err
```
darioAnongba, line 32: I'm not sure but I feel that the error handling here should return an error from the `errors` package and log?

cmd/flags/auth.go
ggarri: this will be handle by the CLI which does not have any errors parsing it won't mind

pkg/auth/auth0.go
ggarri: Moving impl to replace golang-jwt by go-jose

src/auth/authenticator/oicd/jwt.go
ggarri: Done

src/auth/authenticator/oicd/authenticator.go
ggarri: Yes, in case no token is passed Auth validator is skip and deletegate to next authenticator (TLS, API_KEY) to extract the user info. In case none of them are failed user will be obtain the `anonymous` role and whatever that role has permissions for

pkg/auth/auth0.go
ggarri: Replaced by jose.JSONWebKeySet{}.

src/auth/authenticator/oicd/authenticator.go
ggarri: The authenticator component will convert whatever checker error into a UnauhorizedError
```
	claims, err := a.jwtChecker.Check(req.Context(), token)
	if err != nil {
		return nil, errors.UnauthorizedError(err.Error())
	}
	```

File: pkg/auth/jwks.go [hunk 9]
```diff
@@ -0,0 +1,49 @@
+package auth
```
nmvalera, line 1: Is auth the right package for this? Looks like it could be used for other purposes than authentication.

Maybe just `pkg/jwks/jwks.go` or `pkg/utils/jwks.go`

Makefile
```diff
@@ -64,7 +64,7 @@ run-coverage:
 coverage: run-coverage
 	@$(OPEN) build/coverage/coverage.html 2>/dev/null
 
-dev: gobuild
+dev: deps gobuild
```
darioAnongba, line 67: lol you keep adding this and I keep removing it

Makefile
```diff
@@ -76,9 +76,10 @@ down: down-go-quorum down-besu
 
 down-dev:
 	@docker-compose -f ./docker-compose.yml down --volumes --timeout 0
+	@make down-deps
```
darioAnongba, line 79: lol you keep adding this and I keep removing it. I don't want my deps to go down when I take down my qkm. Maybe we can add a new command like `up-qkm` ?

pkg/auth/jwks.go
```diff
@@ -0,0 +1,49 @@
...
+	Kid string   `json:"kid"`
+	X5c []string `json:"x5c"`
+	X5t string   `json:"x5t,omitempty"`
+}
+
+func RetrieveKeySet(ctx context.Context, client *http.Client, authEndpoint string) (*jose.JSONWebKeySet, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", authEndpoint, nil)
+	response, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("call to JWK server failed. %s", err.Error())
```
darioAnongba, line 35: can return err directly

pkg/auth/jwks.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to retrieve keys from JWK server.")
+	}
+
+	keySet := &jose.JSONWebKeySet{}
+	if err := json2.UnmarshalBody(response.Body, keySet); err != nil {
+		return nil, fmt.Errorf("failed to decode response body. %s", err.Error())
```
darioAnongba, line 45: can return err directly

pkg/auth/jwks.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	json2 "github.com/consensys/quorum-key-manager/pkg/json"
+	"gopkg.in/square/go-jose.v2"
+)
+
+const (
```
darioAnongba, line 12: is this used?

pkg/auth/jwks.go
```diff
@@ -0,0 +1,49 @@
...
+	Auth0IssuerServerDomain = "auth0.com"
+)
+
+type JWKsResponse struct {
```
darioAnongba, line 16: Is this still needed?

pkg/tls/certificate/certificate_test.go
```diff
@@ -1,11 +1,9 @@
-package certificate_test
+package certificate
```
darioAnongba, line 1: not sure if this `certificate` package has any purpose? :thinking:

src/auth/authenticator/oicd/authenticator.go
```diff
@@ -0,0 +1,58 @@
...
+func (a Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// Extract Access Token from context
+	token, ok := extractToken(BearerSchema, req.Header.Get("Authorization"))
+	if !ok {
+		return nil, nil
+	}
+
+	claims, err := a.jwtChecker.Check(req.Context(), token)
+	if err != nil {
+		return nil, errors.UnauthorizedError(err.Error())
```
darioAnongba, line 42: shouldn't Unauthorized be returned by the Checker directly? Also, are we not logging?

src/auth/authenticator/oicd/jwt.go
```diff
@@ -0,0 +1,82 @@
...
+	// Parse and validate token injected in context
+	token, err := checker.parser.ParseWithClaims(
+		bearerToken,
+		&Claims{cfg: checker.claimsCfg},
+		checker.keyFunc,
+	)
+	if err != nil {
+		return nil, err
+	} else if !token.Valid {
+		return nil, fmt.Errorf("invalid access token")
```
darioAnongba, line 44: I think that no `fmt.Errorf` should exist under `src` folder. This seems to be an Unauthorized error

src/auth/authenticator/oicd/jwt.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return token.Claims.(*Claims), nil
+}
+
+func (checker *JWTChecker) keyFunc(token *jwt.Token) (interface{}, error) {
+	_, ok := token.Method.(*jwt.SigningMethodRSA)
+	if !ok {
+		return nil, fmt.Errorf("invalid access token signing method")
```
darioAnongba, line 53: same comment on Unauthorized error

src/auth/authenticator/oicd/jwt.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, err
+	} else if !token.Valid {
+		return nil, fmt.Errorf("invalid access token")
+	}
+
+	return token.Claims.(*Claims), nil
+}
+
+func (checker *JWTChecker) keyFunc(token *jwt.Token) (interface{}, error) {
+	_, ok := token.Method.(*jwt.SigningMethodRSA)
```
darioAnongba, line 51: It seems that we also allow SIgningMethodECDSA?

src/auth/authenticator/oicd/jwt.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("invalid access token signing method")
+	}
+
+	for _, cert := range checker.certs {
+		if pubkey, err := tokenAlgoChecker(token.Method.Alg(), cert); err == nil {
+			return pubkey, nil
+		}
+	}
+
+	return nil, fmt.Errorf("unable to find appropriate key in key set.")
```
darioAnongba, line 62: same comment on Unauthorized error

src/auth/authenticator/oicd/jwt.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func tokenAlgoChecker(algo string, cert *x509.Certificate) (interface{}, error) {
```
darioAnongba, line 65: this function should not exist anymore, you can check the code I sent you. You just need to check the `kid` value. These hardcoded values `"RS256", "RS384", "RS512"` are what `token.Method.(*jwt.SigningMethodRSA)` is doing.

Something like this:
```
_, ok := token.Method.(*jwt.SigningMethodRSA)
		if !ok {
			errMessage := "invalid access token signing method"
			return nil, errors.New(errMessage)
		}
		for _, key := range keySet.Keys {
			if token.Header["kid"] == key.KeyID {
				return key.Certificates[0].PublicKey, nil
			}
		}
```
keySet comes from go-jose

src/auth/authenticator/oicd/testutils/jwtGenerator.go
ggarri: it is part of our CMD utils to generate tokens to work with IMO it is useful to provide such a thing and it is not part of our src but pkg what I believe it is good since it does to mix with the app codebase

pkg/auth/jwks.go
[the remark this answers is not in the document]
ggarri: SInce this logic appears to be entirely decouple from our application I put it there

Makefile
ggarri: Actually I was wondering why this was missing I was almost 100% sure I added xD. 

Since we require `deps` to be up for running `make dev` I think we should exact the oppositive when you do `make down`

src/auth/authenticator/oicd/jwt.go
ggarri: Checker errors are all converted into `Unauthorized` on the Authenticator (As you see below). What would it be the reason to have different errors returned by the checker?
```
claims, err := a.jwtChecker.Check(req.Context(), token)
	if err != nil {
		return nil, errors.UnauthorizedError(err.Error())
	}
```

src/auth/authenticator/oicd/jwt.go
ggarri: That was dumb from my side and forgot to remove this from the template auth0 code as I already do the checks on allowed token method on the `tokenAlgoChecker()`

src/auth/authenticator/oicd/jwt.go
ggarri: I don't want me implementation to be so tight to a third party library. What would it be the advantage on using this go-jose?

darioAnongba (approved):

PR #212, opened 2021-07-15
#206 review

dolanor:
Instead of commenting on the already merged #206 PR, and because I also wanted to comment/ask a little bit out of the diffs, just created a PR with comment was easier.

During my analysis, a go test -race discovered some race condition that should be taken into account, I think.
It is not because of this PR though, so we should maybe just add an issue to fix this.
I would make the CI run the tests with -race, and fix the existing race conditions.

My other comments are mostly about error that are ignored, and I wanted to know why. At some point, the error chain is interrupted by a func not returning an error (BaseManager.loadAll()), and then the calling function returns an error again (BaseManager.Start()), which makes us either create error out of the void with a loss of context, or not do anything at all (the current way), which leaves us to debug and dig into code when that happens.

### DIFF
--- a/src/auth/policy/service.go
+++ b/src/auth/policy/service.go
@@ -12,6 +12,7 @@ import (
 	manifest "github.com/consensys/quorum-key-manager/src/manifests/types"
 )
 
+//the: avoid globals?
 var authKinds = []manifest.Kind{
 	GroupKind,
 	Kind,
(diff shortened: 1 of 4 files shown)


darioAnongba:
TBH I think we can close this PR because we are all aware of the issues of the Managers currently. In the future, Managers won't exist and they will be split between a DB (storage of stores and nodes) and connectors (use cases over stores). The concept of Start() and Stop() will disappear from the managers and reading manifests will move to a process in the service layer.

dolanor:
This PR is mostly to comment on what needs a check. I agree to close it, but I think some points still need attention (like error ignoring).
We should at least track it with issues if we agree on the points I made.

dolanor:
as discussed with @ggarri , he agrees with most ideas. I just need to take time to take care of it.

darioAnongba:
Tickets have been created to address the comments of this PR. Closing.

PR #213, opened 2021-07-15
Feature/114 tls auth

6l20:
## PR description

TLS authenticator implementation

## Changelog

- Enables `auth-tls-ca` parameter for KM
- Adds 1 Auth MW
- Compares client certificate
- Accepts identical one
- Reject others
- UserInfo.UserName = CN
- UserInfo.Groups = Org


cmd/flags/auth.go
```diff
@@ -64,11 +59,58 @@ const (
 	authOIDCClaimGroupEnv      = "AUTH_OIDC_CLAIM_GROUPS"
 )
 
+func init() {
```
ggarri, line 62: init methods must be at the beggining of the file by convention if I am not mistaken

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return auth, nil
+}
+
+// Authenticate checks certs and retrieve user Info
+// CN field -> Username
+// Organisation -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract Certificate info from request if any
+	if len(req.TLS.PeerCertificates) > 0 {
```
ggarri, line 35: Do you mind to rewrite the oppositive condition here so it looks like:
```
if len(req.TLS.PeerCertificates) == 0 {
return nil, errors.UnauthorizedError("no cert found in request")
}
```

Actually our linting so be capable to catch this which optmize the improve the code readability
```

File: src/auth/authenticator/tls/authenticator.go [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+		// As mentioned in doc first array element is the leaf
```
ggarri, line 36: This annotation won't mean much in few weeks as we won't know which doc we refer to. IMO it would be better if we look for the actual tecnical reason making this statement true

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,61 @@
...
+				Groups:   clientCert.Subject.Organization,
+				AuthMode: AuthMode,
+			}, nil
+		} else {
+			return nil, errors.UnauthorizedError("certs do not match")
+		}
+
+	}
+
+	return nil, errors.UnauthorizedError("no cert found in request")
```
ggarri, line 60: in case of missing certificate we should ignore this authenticator and let next one to check and in case none of them reject or accept, user will go thought with `annonymous` role

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,61 @@
...
+				break
+			}
+		}
+		if matchingCert {
+			return &types.UserInfo{
+				Username: clientCert.Subject.CommonName,
+				Groups:   clientCert.Subject.Organization,
+				AuthMode: AuthMode,
+			}, nil
+		} else {
```
ggarri, line 54: Unnecesary else statement

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,58 @@
+package tls
```
ggarri, line 1: We are missing the unit test for this

cmd/flags/auth.go
```diff
@@ -108,26 +154,40 @@ Environment variable: %q`, authOIDCCACertFileEnv)
...
+		vipr.GetString(authOIDCClaimGroupViperKey), certsOIDC...)
+
+	// TLS part
+	certsTLS := []*x509.Certificate{}
+
+	fileCertTLS, err := tlsAuthClientCertificate(vipr)
+	if err != nil {
+		return nil, err
+	} else if fileCertTLS != nil {
+		certsTLS = append(certsTLS, fileCertTLS)
```
ggarri, line 184: Why you need to append into an array instead of doing the assigment at the initialization?

src/auth/authenticator/tls/authenticator.go
6l20: This one is the PeerCertificate doc, should not move too much I think. Anyway mentioning the doc here is not clear. Ok for me

src/auth/authenticator/tls/authenticator.go
6l20: Ok

cmd/flags/auth.go
6l20: Ok

src/auth/authenticator/tls/authenticator.go
6l20: Ok, I will return anonymous

src/auth/authenticator/tls/authenticator.go
6l20: Yes saw that afterwards

cmd/flags/auth.go
6l20: Well I guess I just did copy/paste the above code. Better without indeed.

src/auth/authenticator/tls/authenticator.go
ggarri: Actually you should return nil nil, and the authenticathor middleware will take care of appending anonymous role to the user request

cmd/flags/auth.go
```diff
@@ -108,26 +154,36 @@ Environment variable: %q`, authOIDCCACertFileEnv)
...
+	} else if fileCertOIDC != nil {
+		certsOIDC = append(certsOIDC, fileCertOIDC)
 	}
 
 	issuerCerts, err := issuerCertificates(vipr)
 	if err != nil {
 		return nil, err
 	} else if issuerCerts != nil {
-		certs = append(certs, issuerCerts...)
+		certsOIDC = append(certsOIDC, issuerCerts...)
```
ggarri, line 171: Minor: Still doing append when it is not needed

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	// first array element is the leaf
+	clientCert := *req.TLS.PeerCertificates[0]
+	// check this cert matches authenticator provided one
+	// using strict comparison
+	var matchingCert bool
+
+	for _, authCert := range authenticator.Certificates {
+		if bytes.Equal(clientCert.Raw, authCert.Raw) {
+			matchingCert = true
```
ggarri, line 44: Could return directly here the userInfo

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,56 @@
...
+	return auth, nil
+}
+
+// Authenticate checks certs and retrieve user Info
+// CN field -> Username
+// Organization -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract Certificate info from request if any
+	if len(req.TLS.PeerCertificates) == 0 {
+		return types.AnonymousUser, nil
```
ggarri, line 34: you should return nil, nil and then AuthenticathorMid will take care of appending `AnnonymousUser` groups to the UserInfo attached in the context

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	// first array element is the leaf
+	clientCert := *req.TLS.PeerCertificates[0]
+	// check this cert matches authenticator provided one
+	// using strict comparison
+	var matchingCert bool
+
+	for _, authCert := range authenticator.Certificates {
+		if bytes.Equal(clientCert.Raw, authCert.Raw) {
```
ggarri, line 43: You could have used certificate Equal to match if boths certs are equal:
```
func (c *Certificate) Equal(other *Certificate) bool {
	if c == nil || other == nil {
		return c == other
	}
	return bytes.Equal(c.Raw, other.Raw)
}
```

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,56 @@
...
+// Authenticate checks certs and retrieve user Info
+// CN field -> Username
+// Organization -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract Certificate info from request if any
+	if len(req.TLS.PeerCertificates) == 0 {
+		return types.AnonymousUser, nil
+	}
+	// first array element is the leaf
+	clientCert := *req.TLS.PeerCertificates[0]
```
ggarri, line 37: Could you confirm with @nmvalera if the expected check should only refer to the cert leaf and not to the entire VerifiedChain of certificates of the request.

```
// VerifiedChains is a list of one or more chains where the first element is
	// PeerCertificates[0] and the last element is from Config.RootCAs (on the
	// client side) or Config.ClientCAs (on the server side).
	//
	// On the client side, it's set if Config.InsecureSkipVerify is false. On
	// the server side, it's set if Config.ClientAuth is VerifyClientCertIfGiven
	// (and the peer provided a certificate) or RequireAndVerifyClientCert.
	VerifiedChains [][]*x509.Certificate
```

src/auth/authenticator/tls/authenticator.go
6l20: Yes, I did not spot that one

cmd/flags/auth.go
6l20: This part is the OIDC one, seems needs

src/auth/authenticator/tls/authenticator.go
6l20: Ok, I have changed this. But in that case I am not sure that the duty of the upper layer is to guess what user it is dealing with. Responsability of Authenticate is to return useInfo. Here we should probably consider always returning a proper user instance and let the upper layer deal with it.

darioAnongba (approved):
Ok for me as long as you check that it works fine and also @ggarri about the leaf cert

pkg/app/app.go
```diff
@@ -218,19 +226,43 @@ func (app *App) startServer() {
...
-		app.logger.Info("started API server", "addr", app.server.Addr)
-		app.errors <- app.server.ListenAndServe()
+		app.logger.Info("starting API server", "addr", app.server.Addr)
+		apiErr := app.server.ListenAndServe()
+		if apiErr == nil {
+			app.logger.Debug("started API server successfully", "addr", app.server.Addr)
+		} else {
+			app.logger.WithError(apiErr).Info("failed to start API server")
+		}
+		app.errors <- apiErr
```
darioAnongba, line 240: good stuff

pkg/app/app.go
```diff
@@ -218,19 +226,43 @@ func (app *App) startServer() {
...
 	}()
 
 	go func() {
 		app.logger.Info("started Health server", "addr", app.healthz.Addr)
-		app.errors <- app.healthz.ListenAndServe()
+		healthErr := app.healthz.ListenAndServe()
+		if healthErr == nil {
+			app.logger.Debug("started Health server successfully", "addr", app.server.Addr)
+		} else {
+			app.logger.WithError(healthErr).Info("failed to start API TLS server")
```
darioAnongba, line 260: wrong message and should be `Error`

pkg/app/app.go
```diff
@@ -218,19 +226,43 @@ func (app *App) startServer() {
...
+		app.errors <- apiErr
+	}()
+
+	go func() {
+		app.logger.Info("starting API TLS server", "addr", app.tlsServer.Addr, "cert", app.cfg.HTTP.TLSCert, "key", app.cfg.HTTP.TLSKey)
+		tlsErr := app.tlsServer.ListenAndServeTLS(app.cfg.HTTP.TLSCert, app.cfg.HTTP.TLSKey)
+		if tlsErr == nil {
+			app.logger.Debug("started API TLS server successfully", "addr", app.server.Addr)
+		} else {
+			app.logger.WithError(tlsErr).Info("failed to start API TLS server")
```
darioAnongba, line 249: should be `Error`

pkg/app/app.go
```diff
@@ -245,7 +277,12 @@ func (app *App) stopServer(ctx context.Context) error {
 		return err
 	}
 
-	app.logger.Info("servers (API and Health) gracefully shut down")
+	if err := app.tlsServer.Shutdown(ctx); err != nil {
+		app.logger.WithError(err).Error("tls api server could not shut down")
+		return err
+	}
+
+	app.logger.Info("servers (API, TLS and Health) gracefully shut down")
```
darioAnongba, line 285: nice

darioAnongba (approved):

src/auth/authenticator/tls/authenticator.go
ggarri: that would work in case of having a single auth alternatives. In meanwhile we are supporting several ones it wont work since we need a parent auth components to decide the final outcome. What else would you propose here?

src/auth/authenticator/tls/authenticator.go
ggarri: @6l20 bumping this before merging

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,51 @@
...
+			return &types.UserInfo{
+				Username: clientCert.Subject.CommonName,
+				Groups:   clientCert.Subject.Organization,
+				AuthMode: AuthMode,
+			}, nil
+
+		}
+	}
+
+	return nil, errors.UnauthorizedError("no matching cert found")
```
ggarri, line 50: As far as I see we should diferenciate between using a standard tls connection to protect client to server communications, which will be validated by your http server, and using a tls certificate to authenticate your connection including username and groups. That in why I believe it is important to dive into the provided chain of certs in this middleware.

 In addition I think we should not reject in case of invalid certificate because of those reasons. wdyt @nmvalera

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,51 @@
...
+
+	return auth, nil
+}
+
+// Authenticate checks certs and retrieve user Info
+// CN field -> Username
+// Organization -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract Certificate info from request if any
+	if len(req.TLS.PeerCertificates) == 0 {
```
nmvalera, line 32: Should we add if `req.TLS != nil`? Looks like this may `panic` if TLS is off

src/auth/authenticator/tls/authenticator.go
nmvalera: Actually. 

Looks like Kubernetes extract from the leaf only: https://github.com/kubernetes/kubernetes/blob/c224ca7674445ec98a2e8876daf53b9846cd3ae1/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go#L164

That uses `CommonUserNameConversion` from package `x509` : https://pkg.go.dev/k8s.io/Kubernetes@v1.13.13-beta.0/staging/src/k8s.io/apiserver/pkg/authentication/request/x509#pkg-variables

src/auth/authenticator/tls/authenticator.go
nmvalera: c.f. Kubernetes: https://github.com/kubernetes/kubernetes/blob/c224ca7674445ec98a2e8876daf53b9846cd3ae1/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go#L134

6l20: This was indeed an issue with http

src/auth/authenticator/tls/authenticator.go
6l20: Leaf only is considered here, moreover it is the only part of the cert we may have the hands on before deployment

src/auth/authenticator/tls/authenticator.go
6l20: As of today certificate matching is no longer in place.
Every valid and provided cert will be treated as a correct authentication

pkg/app/app.go
6l20: Ok, fixed

PR #214, opened 2021-07-15
Alias db backend

### LINKED ISSUE #146: Create alias service "backend" allowing to perform CRUD operations for aliases
#### Summary

Create backend to store aliases to database

- An alias is attached to one and only one registry
- An alias holds a value which can be any JSON object or an array
- Each alias as a type (in string format) passed by the users

#### Acceptance Criteria

- [ ] Create Postgres database schema
- [ ] Add migration scripts for the database
- [ ] Add configuration flags for the database
- [ ] Add method to create and store an alias
- [ ] Add method to load an alias
- [ ] Add method to delete an alias
- [ ] Add method to list aliases attached to a registry
- [ ] Add method to update an alias
- [ ] Add unit-test

dolanor:
## PR description

Add DB backend for aliases

## Fixed Issue(s)
- fixes #146


Makefile
```diff
@@ -49,8 +51,8 @@ down-deps: postgres-down hashicorp-down down-networks
 run-acceptance:
 	@go test -v -tags acceptance -count=1 ./tests/acceptance
 
-run-e2e:
-	@go test -v -tags e2e -count=1 ./tests/e2e
+run-e2e: deps
```
darioAnongba, line 54: I wouldn't add this step before e2e. e2e are independent on the start of the dependencies.

Makefile
```diff
@@ -49,8 +51,8 @@ down-deps: postgres-down hashicorp-down down-networks
...
+	@go test -v -tags e2e -count=1 ./tests/e2e ./src/aliases/store
```
darioAnongba, line 55: why do you need `./src/aliases/store` here?

deps/docker-compose.yml
```diff
@@ -3,10 +3,11 @@ version: '3.7'
 x-container-common: &container-common
   restart: ${CONTAINER_RESTART-on-failure}
   networks:
-    - qkm
+    - postgres
```
darioAnongba, line 6: qkm is correct

deps/docker-compose.yml
```diff
@@ -21,10 +22,28 @@ services:
     ports:
       - 5432
 
+  pgadmin:
```
darioAnongba, line 25: please remove everything related to pgadmin from this docker-compose. pgadmin has already been added to the `docker-compose-tools.yml` in this folder. We should not mix tools and dependencies.

docker-compose.yml
```diff
@@ -27,17 +27,17 @@ x-container-common: &container-common
     - quorum
     - besu
     - hashicorp
-    - qkm
+    - postgres
```
darioAnongba, line 30: postgres is a mandatory dependency of the qkm, so qkm is correct here.

docker-compose.yml
```diff
@@ -27,17 +27,17 @@ x-container-common: &container-common
...
 
-x-qkm-common: &qkm-common
+x-postgres-common: &postgres-common
```
darioAnongba, line 32: same qkm remark.

docker-compose.yml
```diff
@@ -27,17 +27,17 @@ x-container-common: &container-common
...
   DB_HOST: postgres
 
 services:
   key-manager:
     <<: *container-common
     environment:
       <<: *default-variables
-      <<: *qkm-common
+      <<: *postgres-common
```
darioAnongba, line 40: same qkm remark.

docker-compose.yml
```diff
@@ -74,9 +74,9 @@ networks:
   hashicorp:
     external:
       name: hashicorp
-  qkm:
+  postgres:
```
darioAnongba, line 77: same qkm remark.

pkg/app/app.go
```diff
@@ -161,7 +162,7 @@ func (app *App) RegisterService(srv interface{}) error {
 	if rSrv, ok := srv.(common.Runnable); ok {
 		app.services = append(app.services, reflect.ValueOf(rSrv))
 	} else {
-		errMessage := "registered service is not a runnable"
+		errMessage := fmt.Sprintf("registered service is not a runnable: %T", srv)
```
darioAnongba, line 165: parameters are not added in the error message, they are added in the typed data of the log. So here what you want is:
```
errMessage := "registered service is not a runnable"
app.logger.With("service", srv).Error(errMessage)
return errors.ConfigError(errMessage)
```
That is called structured logging

src/aliases/store/migrations/001_add_aliases.up.sql
```diff
@@ -0,0 +1,18 @@
+begin;
+
+create type alias_kind as enum (
```
darioAnongba, line 3: I find this weird... How can a type be 2 things at the same time? An array or a string. This breaks good practices.

src/aliases/manager/service.go
```diff
@@ -0,0 +1,25 @@
+package aliasservice
```
darioAnongba, line 1: It seems you don't need this `manager` folder in this PR

src/aliases/store/store.go
```diff
@@ -0,0 +1,50 @@
+package aliasstore
+
+import "context"
+
+//go:generate mockgen -source=store.go -destination=mock/store.go -package=mock
+
+// Store handles the alias storing.
+type Store interface {
```
darioAnongba, line 8: `Store` is your `Alias` agent that should be appended to the `database.Database` interface in the same way as `ETH1Accounts`, `Keys` and `Secrets`. We can have a call to peer on this to have a consistent way of defining agents.

src/aliases/store/store.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// DeleteRegistry deletes a registry, with all the aliases it contained.
+	DeleteRegistry(ctx context.Context, registry RegistryName) error
+}
+
+// Alias allows the user to associates a RegistryName + a Key to 1 or more public keys stored
+// in Value. The Value has 2 formats:
+// - a JSON string if AliasKind is an AliasKindString.
+// - a JSON array of strings if AliasKind is an AliasKindArray.
+type Alias struct {
```
darioAnongba, line 29: `Alias` is your model, it should go under `aliases/store/models/alias.go`

src/aliases/store/store.go
```diff
@@ -0,0 +1,50 @@
...
+	tableName struct{} `pg:"aliases"` // nolint:unused,structcheck // reason
+
+	Key          AliasKey     `pg:",pk"`
+	RegistryName RegistryName `pg:",pk"`
```
darioAnongba, line 33: From what I see, it seems you have 2 entities in your data representation:
- A `Registry`, which is a list of aliases, a registry has a `name`
- An `Alias`, which is a resource that gets added to a registry.

So apparently you need 2 models, one `Registry` and one `Alias` where the alias is linked to a registry by name. Having a unique index on `(alias, registry)` is also fine but inferior representation of the data. 

PS: You don't seem to be creating registries so maybe I'm wrong on this one and the registry is something that we do not control?

src/app.go
```diff
@@ -66,6 +67,11 @@ func New(cfg *Config, logger log.Logger) (*app.App, error) {
 		return nil, err
 	}
 
+	err = aliases.RegisterService(a, postgresClient)
```
darioAnongba, line 70: Seems you can skip registering service in this PR as you only implemented the data layer.

src/infra/postgres/client/client.go
```diff
@@ -120,3 +132,24 @@ func (c PostgresClient) RunInTransaction(ctx context.Context, persist func(clien
...
+		return err
+	}
+	return nil
+}
+
+func (c *PostgresClient) Ping(ctx context.Context) error {
+	return c.pgdb.Ping(ctx)
+}
+
+func (c *PostgresClient) CreateTable(ctx context.Context, model interface{}) error {
```
darioAnongba, line 149: This cannot exist in this client as it should never be needed in the app.

src/infra/postgres/client/client.go
```diff
@@ -120,3 +132,24 @@ func (c PostgresClient) RunInTransaction(ctx context.Context, persist func(clien
...
+	return c.db.ModelContext(ctx, model).CreateTable(nil)
+}
+
+func (c *PostgresClient) DropTable(ctx context.Context, model interface{}) error {
+	return c.db.ModelContext(ctx, model).DropTable(nil)
```
darioAnongba, line 154: This cannot exist in this client as it should never be needed in the app.

src/infra/postgres/client/client.go
```diff
@@ -120,3 +132,24 @@ func (c PostgresClient) RunInTransaction(ctx context.Context, persist func(clien
...
+func (c *PostgresClient) SelectMany(ctx context.Context, model, dst interface{}, condition string, params ...interface{}) error {
+	q := c.db.ModelContext(ctx, model)
+	err := q.Where(condition, params...).Select(dst)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *PostgresClient) Ping(ctx context.Context) error {
```
darioAnongba, line 145: I suppose you need this in the migration process? If so, it should not appear here.

src/infra/postgres/client/client.go
```diff
@@ -3,15 +3,17 @@ package client
...
 	"github.com/consensys/quorum-key-manager/src/infra/postgres"
-	"github.com/go-pg/pg/v10"
 )
 
 type PostgresClient struct {
-	cfg *Config
-	db  orm.DB
+	cfg  *Config
+	db   orm.DB
+	pgdb *pg.DB
```
darioAnongba, line 16: pgdb = db where db is the interface and pgdb is the struct so here you should not need pgdb

src/infra/postgres/postgres.go
```diff
@@ -16,4 +16,10 @@ type Client interface {
 	DeletePK(ctx context.Context, model ...interface{}) error
 	ForceDeletePK(ctx context.Context, model ...interface{}) error
 	RunInTransaction(ctx context.Context, persist func(client Client) error) error
+
+	SelectMany(ctx context.Context, model interface{}, dst interface{}, condition string, params ...interface{}) error
+
+	Ping(ctx context.Context) error
+	CreateTable(ctx context.Context, model interface{}) error
+	DropTable(ctx context.Context, model interface{}) error
```
darioAnongba, line 24: Same comment about this should not appear here as it's not used by the app.

src/infra/postgres/postgres.go
```diff
@@ -16,4 +16,10 @@ type Client interface {
 	DeletePK(ctx context.Context, model ...interface{}) error
 	ForceDeletePK(ctx context.Context, model ...interface{}) error
 	RunInTransaction(ctx context.Context, persist func(client Client) error) error
+
+	SelectMany(ctx context.Context, model interface{}, dst interface{}, condition string, params ...interface{}) error
+
+	Ping(ctx context.Context) error
+	CreateTable(ctx context.Context, model interface{}) error
```
darioAnongba, line 23: Same comment about this should not appear here as it's not used by the app.

src/infra/postgres/postgres.go
```diff
@@ -16,4 +16,10 @@ type Client interface {
 	DeletePK(ctx context.Context, model ...interface{}) error
 	ForceDeletePK(ctx context.Context, model ...interface{}) error
 	RunInTransaction(ctx context.Context, persist func(client Client) error) error
+
+	SelectMany(ctx context.Context, model interface{}, dst interface{}, condition string, params ...interface{}) error
+
+	Ping(ctx context.Context) error
```
darioAnongba, line 22: Same comment about this should not appear here as it's not used by the app.

tests/e2e/postgres_test.go
```diff
@@ -0,0 +1,166 @@
+// +build e2e
```
darioAnongba, line 1: this can be removed as it's not an e2e test. e2e tests by definition do not test dependencies so having an e2e test called `TestPostgres` is a contradiction by itself.

Makefile
dolanor: It still needs running dependencies. How are we supposed to run them? Isn't it with `make deps`?

Makefile
dolanor: A wrong assumption of what the e2e were, and some tests I ran with real dependencies were in that folder. It moved since, but I forgot to update this.
```suggestion
	@go test -v -tags e2e -count=1 ./tests/e2e
```

src/aliases/store/store.go
dolanor: We control the registry, it is just not used in the DB as it is because we don't really need to. It is just to scope aliases from other registries.
I'll split the entities and models then.

src/infra/postgres/client/client.go
dolanor: I needed access to be able to ping the DB, `orm.DB` doesn't offer this functionality.

docker-compose.yml
dolanor: ok

pkg/app/app.go
dolanor: You're right, I could have done that.
Some [other error string contradict that](https://github.com/ConsenSys/quorum-key-manager/blob/50e060562dca6060270456dfb1752459621c8d67/pkg/app/app.go#L102), so I just inspired myself from what was there. Also, nothing is documented for the logger interface, which makes more digging. We could figure out kind of easily that it's zap behind, but it's more and more reference to dig.
Also, I added that because it helped me debug the app. I think it should be part of another PR.
I'm gonna remove it and make another PR.

src/aliases/manager/service.go
dolanor: I guess you're right. I shouldn't have commited it, it was just to confirm that my code was loading well.
I'll remove it and the lines from `src/app.go`

src/app.go
dolanor: yes, [I agree](https://github.com/ConsenSys/quorum-key-manager/pull/214#discussion_r679090238)

tests/e2e/postgres_test.go
dolanor: Yes, misunderstanding about what is the difference between e2e and acceptance tests.
And maybe this test has no reason to exist at all? Or another category of tests that make sure that our interfaces works with the real backend system before using for mocking.

dolanor:
I fixed everything that was obviously wrong, for the rest, I agree we need to peer on it and find a common approach.

src/infra/postgres/client/client.go
dolanor: It will be used for healthcheck -> rename to Healthcheck.

dolanor: check in orchestrate

src/infra/postgres/client/client.go
dolanor: Finally I removed it with the tests not using dockertest anymore.

src/infra/postgres/postgres.go
dolanor: We decided to keep it in the review finally until we use real migrations even for tests.

src/infra/postgres/postgres.go
dolanor: We decided to keep it in the review finally until we use real migrations even for tests.

src/infra/postgres/postgres.go
dolanor: Finally I removed it with the tests not using dockertest anymore.

tests/e2e/postgres_test.go
dolanor: This has been removed entirely and will be added back once we refactor the acceptance/integration tests into 1 test runner.

src/infra/postgres/client/client.go
dolanor: We decided to keep it in the review finally until we use real migrations even for tests.

src/infra/postgres/client/client.go
dolanor: We decided to keep it in the review finally until we use real migrations even for tests.

src/aliases/store/store.go
dolanor: I'll move it there, but I would like to discuss that at our next meeting.

src/aliases/store/store.go
dolanor: We decided to keep the database a direct dependency of each service, so no global Database interface should group every service datastore together.

tests/e2e/postgres_test.go
dolanor: We decided to include in the acceptance (that will be renamed to integration), I'll integrate it when we make the acceptance test run only 1 test environment. (currently, for my test, I created a parallel test env, we need to merge)

src/infra/postgres/client/client.go
dolanor: Removed for now, it's handled in another way with the acceptance tests.

src/aliases/store/migrations/001_add_aliases.up.sql
dolanor: This is removed entirely. We'll try the idea of just storing a JSON array of string for the keys.

pkg/app/app.go
dolanor: referenced here: https://github.com/ConsenSys/quorum-key-manager/issues/239

dolanor:
I still have https://app.circleci.com/pipelines/github/ConsenSys/quorum-key-manager/935/workflows/92c9d31d-5c88-4286-ac16-46336f58988e/jobs/3782?invite=true#step-112-36
But it seems it's completely unrelated.

Makefile
darioAnongba: running the e2e is independent of setting your environment. You can do one `make deps` but then run 10x `make run-e2e`. These 2 processes are not linked. The e2e expect that the env is already up.

Makefile
```diff
@@ -30,14 +30,16 @@ networks:
 	@docker network create --driver=bridge hashicorp || true
 	@docker network create --driver=bridge --subnet=172.16.237.0/24 besu || true
 	@docker network create --driver=bridge --subnet=172.16.238.0/24 quorum || true
+	@docker network create --driver=bridge qkm || true
```
darioAnongba, line 33: @ggarri If I remember correctly you told me to not do that for postgres and the qkm network

deps/docker-compose.yml
```diff
@@ -24,7 +25,10 @@ services:
...
   db-data-api:
     driver: local
+  state-pgadmin:
+    driver: local
 
 networks:
   qkm:
-    driver: bridge
\ No newline at end of file
+    external:
```
darioAnongba, line 33: please remove all the changes of this file

docker-compose.yml
```diff
@@ -76,7 +76,7 @@ networks:
       name: hashicorp
   qkm:
     external:
-      name: deps_qkm
+      name: qkm
```
darioAnongba, line 79: same here, no changes needed for this PR

src/aliases/store/postgres/postgres.go
```diff
@@ -0,0 +1,58 @@
+package aliaspg
```
darioAnongba, line 1: You need to follow the same structure as the `stores` database folder, you don't need a folder called `database` and your root is `store` (because the data layer of aliases is only the DB):

- store
- - database.go
- - postgres
- - - alias.go
- - - database.go
- - models
- - - alias.go
- - mock
- - - database.go

Even if you only have one Repository/Data agent for now, it's better to have a `Database` interface that has an `Alias()` function for consistency and scalability.

src/aliases/store/postgres/postgres.go
```diff
@@ -0,0 +1,58 @@
...
+
+	models "github.com/consensys/quorum-key-manager/src/aliases/models"
+	aliasstore "github.com/consensys/quorum-key-manager/src/aliases/store"
+	"github.com/consensys/quorum-key-manager/src/infra/postgres"
+)
+
+var _ aliasstore.Store = &Store{}
+
+// Store stores the alias data in a postgres DB.
+type Store struct {
```
darioAnongba, line 15: This is your `alias.go` and this is called `Alias` or `AliasRepository` or similar

src/aliases/store/store.go
darioAnongba: This is your `Alias` interface. Because of the refactor, you don't need to add `Alias` to all your function names, they can just be `Create, Get, etc.`

src/infra/postgres/client/client.go
```diff
@@ -80,19 +84,25 @@ func (c *PostgresClient) SelectDeleted(ctx context.Context, model ...interface{}
 }
 
 func (c *PostgresClient) UpdatePK(ctx context.Context, model ...interface{}) error {
-	_, err := c.db.ModelContext(ctx, model...).WherePK().Update()
+	res, err := c.db.ModelContext(ctx, model...).WherePK().Update()
 	if err != nil {
 		return parseErrorResponse(err)
 	}
+	if res.RowsAffected() != 1 {
```
darioAnongba, line 91: An update operation does not return "not found" if the update doesn't update anything. If the update does not update anything, it's still a success

src/infra/postgres/client/client.go
```diff
@@ -80,19 +84,25 @@ func (c *PostgresClient) SelectDeleted(ctx context.Context, model ...interface{}
...
 	return nil
 }
 
 func (c *PostgresClient) DeletePK(ctx context.Context, model ...interface{}) error {
-	_, err := c.db.ModelContext(ctx, model...).WherePK().Delete()
+	res, err := c.db.ModelContext(ctx, model...).WherePK().Delete()
 	if err != nil {
 		return parseErrorResponse(err)
 	}
+	if res.RowsAffected() != 1 {
```
darioAnongba, line 103: I doubt you need this, if the data is not found, you should get a `err == pg.ErrNoRows`. The data cannot be found but nor deleted, otherwise you get an error. You can simply remove this `if` IMO.

src/infra/postgres/client/client.go
```diff
@@ -45,6 +46,9 @@ func (c *PostgresClient) Insert(ctx context.Context, model ...interface{}) error
 
 func (c *PostgresClient) SelectPK(ctx context.Context, model ...interface{}) error {
 	err := c.db.ModelContext(ctx, model...).WherePK().Select()
+	if err == pg.ErrNoRows {
```
darioAnongba, line 49: You can remove this `if` statement as well. This is done by the `parseErrorResponse` function

tests/acceptance/alias_store_test.go
```diff
@@ -0,0 +1,170 @@
...
+	s := new(aliasStoreTestSuite)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	sig := common.NewSignalListener(func(signal os.Signal) {
+		cancel()
+	})
+	defer sig.Close()
+
+	var err error
+	s.env, err = NewIntegrationEnvironment(ctx)
```
darioAnongba, line 32: You didn't do what we discussed ? Are you planning on doing it in a different PR? IMO it's a really quick change because here you are basically duplicating everything by creating a new `IntegrationEnvironment`.

Makefile
ggarri: Actually he is right, we need to init a `qkm` network to attach deps like postgres. What you did was the oppostive, you added a `postgres` network what didn't make sense imo. In addition to this I think we could remove `hashicorp` bridge an attach it also to `qkm` network, wdyt @dolanor ?

docker-compose.yml
ggarri: this is fine

src/aliases/models/alias.go
```diff
@@ -0,0 +1,20 @@
+package aliasmodels
+
+// Alias allows the user to associates a RegistryName + a Key to 1 or more public keys stored
+// in Value. The Value has 2 formats:
+// - a JSON string if AliasKind is an AliasKindString.
+// - a JSON array of strings if AliasKind is an AliasKindArray.
+type Alias struct {
+	tableName struct{} `pg:"aliases"` // nolint:unused,structcheck // reason
+
+	Key          AliasKey     `pg:",pk"`
```
ggarri, line 10: This model does not correlate to your PG table definition where pk is the a serial id, what is the reason for this?

src/aliases/store/migrations/001_add_aliases.up.sql
```diff
@@ -0,0 +1,12 @@
+begin;
+
+create table if not exists aliases (
+	id serial primary key,
+	key text,
+	registry_name text,
+	value json
+);
+create unique index on aliases (registry_name, key);
+create index on aliases (registry_name);
```
ggarri, line 10: This index won't speed up your queries. Since `registry_name` is the first part of the other unique index. If you want I can explain more in detail what is that.

src/infra/postgres/client/client.go
```diff
@@ -120,3 +130,20 @@ func (c PostgresClient) RunInTransaction(ctx context.Context, persist func(clien
 
 	return c.db.(*pg.DB).RunInTransaction(ctx, persistFunc)
 }
+
+func (c *PostgresClient) SelectMany(ctx context.Context, model, dst interface{}, condition string, params ...interface{}) error {
```
ggarri, line 134: More than a `SelectMany` I would rather name this SelectWithQuery with the following impl:
```
func (c *PostgresClient) SelectWithQuery(ctx context.Context, model interface{}, *pg.Query) {
        q := c.db.QueryContext(ctx, model, query).Select()
	if err != nil {
		return err
	}
	return nil
}
```

Do you think that can make the work?

src/infra/postgres/client/client.go
ggarri: Sorry, dO you mind the to detail why you need this? Our migration are tested as part of the integration test I don't see a place for this either

src/infra/postgres/postgres.go
ggarri: I cannot see reason for this either, sorry :)

src/infra/postgres/postgres.go
ggarri: I cannot see reason for this either, sorry :)

Makefile
dolanor: From what I understand, we have 3 types of dependencies:
- mandatory upstream dependencies, like postgres: without it, QKM shouldn't start
- optional upstream dependencies, like hashicorp vault, AKM, AWS KMS: we need them only if the configuration requires it to.
- downstream dependencies: besu and quorum which are users of QKM.

For mandatory, we need the network, I would say it could even be part of the default docker-compose.yml since without it, QKM shouldn't start.
For optional, they could have their own independant/external network, and given the config, they will be requested or not.
For downstream, they clearly need their own network, they are users of QKM. QKM shouldn't even know about them.

So I'd suggest to put it directly in the default QKM docker-compose.yml and the network would start and die when we stop the QKM. (there is no point of having a solo postgres running without a qkm attached to it)

dolanor: And what I understand of the ownership of docker-compose.yml file
- [./docker-compose.yml](https://github.com/ConsenSys/quorum-key-manager/blob/0a286f9d66f357707eea64326114ff86d7ecc382/docker-compose.yml) is for QKM.
- [./deps/docker-compose.yml](https://github.com/ConsenSys/quorum-key-manager/blob/0a286f9d66f357707eea64326114ff86d7ecc382/deps/docker-compose.yml) is for QKM's mandatory dependencies.- [./deps/docker-compose.yml](https://github.com/ConsenSys/quorum-key-manager/blob/0a286f9d66f357707eea64326114ff86d7ecc382/deps/docker-compose-tools.yml) is for QKM mandatory dependencies' tools.
- [./deps/hashicorp/docker-compose.yml](https://github.com/ConsenSys/quorum-key-manager/blob/0a286f9d66f357707eea64326114ff86d7ecc382/deps/hashicorp/docker-compose.yml) is for QKM optional dependencies
- `./deps/{besu,quorum}/docker-compose.yml` is for QKM downstream dependencies (which are users of QKM)

src/infra/postgres/client/client.go
dolanor: It couples the Store interface to the `go-pg` library, which would break the point of having the interface in my opinion.
I could rename the func, though. I'm not very happy with the naming myself.
Also, I think Dario created another func with many rows return. I need to check it and update my code.

deps/docker-compose.yml
dolanor: ok

tests/acceptance/alias_store_test.go
dolanor: The idea was to do it in another PR.
I actually started putting it in the existing test runner, but it was putting it under a StoreTest, which didn't make sense since alias is not related to any Store (AKV, AWS KMS, Hashicorp Vault).
So it's not really a quick change because we need to create a higher level test runner that would run the Store on one side, and the alias tests on the other side.

src/aliases/models/alias.go
dolanor: forgot to fix the migration script.

It is fixed now.

src/aliases/store/migrations/001_add_aliases.up.sql
dolanor: got it.

src/infra/postgres/client/client.go
dolanor: I renamed it `SelectWhere` and reduced the number of args, without depending on `pg` specific types.

src/infra/postgres/postgres.go
dolanor: Removed it as it is not used by anything anymore.

Makefile
dolanor: I got it, but we should improve that workflow and the documentation about it.

src/aliases/store/postgres/postgres.go
dolanor: I've organized as presented here.
I'd like to discuss the way we create the Database interface, I think there would be a better way. I'll add it to the document of our common approach.

src/aliases/store/postgres/postgres.go
dolanor: ok

docker-compose.yml
dolanor: darioAnongba is right, I forgot the docker-compose up will prepend the project/dir name.

tests/acceptance/alias_store_test.go
dolanor: https://github.com/ConsenSys/quorum-key-manager/issues/249 will fix it

Makefile
dolanor: I'll put it in the root ./docker-compose.yml file after we'll discuss about it.

src/infra/postgres/client/config.go
```diff
@@ -16,18 +17,18 @@ const (
...
-	Password          string
-	Database          string
-	PoolSize          int
-	PoolTimeout       time.Duration
-	DialTimeout       time.Duration
-	KeepAliveInterval time.Duration
-	TLS               *tls.Option
-	ApplicationName   string
-	SSLMode           string
+	Host              string        `json:"host"`
```
ggarri, line 20: Why do you need your config to be mapping a json object? are you doing any usage of it?

src/infra/postgres/postgres.go
```diff
@@ -12,6 +12,7 @@ type Client interface {
 	SelectDeletedPK(ctx context.Context, model ...interface{}) error
 	Select(ctx context.Context, model ...interface{}) error
 	SelectDeleted(ctx context.Context, model ...interface{}) error
+	SelectWhere(ctx context.Context, model interface{}, condition string, params ...interface{}) error
```
ggarri, line 15: Minor: I would rather rename `condition` by `where` just for correlation with the method name

src/aliases/store/postgres/alias.go
```diff
@@ -0,0 +1,58 @@
...
+	return nil
+}
+
+func (s *Alias) GetAlias(ctx context.Context, registryName models.RegistryName, aliasKey models.AliasKey) (*models.Alias, error) {
+	a := models.Alias{Key: aliasKey, RegistryName: registryName}
+	err := s.pgClient.SelectPK(ctx, &a)
+	return &a, err
+}
+
+func (s *Alias) UpdateAlias(ctx context.Context, registryName models.RegistryName, alias models.Alias) error {
```
ggarri, line 40: For consistency, I would suggest to return also the the update object as we do in other DB implementations (see below). I also that is a more standard practice
```
Update(ctx context.Context, account *entities.ETH1Account) (*entities.ETH1Account, error) {
```

src/aliases/store/postgres/alias.go
```diff
@@ -0,0 +1,58 @@
...
+	pgClient postgres.Client
+}
+
+func NewAlias(pgClient postgres.Client) *Alias {
+	return &Alias{
+		pgClient: pgClient,
+	}
+}
+
+func (s *Alias) CreateAlias(ctx context.Context, registryName models.RegistryName, alias models.Alias) error {
```
ggarri, line 25: You are missing the returned object in the create function
```
func (ea *ETH1Accounts) Add(ctx context.Context, account *entities.ETH1Account) (*entities.ETH1Account, error) {
```

src/aliases/store/postgres/alias.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var _ aliasstore.Alias = &Alias{}
+
+// Alias stores the alias data in a postgres DB.
+type Alias struct {
+	pgClient postgres.Client
+}
+
+func NewAlias(pgClient postgres.Client) *Alias {
```
ggarri, line 19: As we design it models are never exposed outside of this layer and we are using entities as our business object representation and as DTO (data transfer object) so every layer of our project communicates using them and each layer might have its internal respresentation to communicate with the outside world (DB, KeyVault, Kafka....).

Doing that we prevent our Business component, such as our `Connectors` to be impacted by changes in our DB model

src/infra/postgres/client/config.go
dolanor: I was using it in the `TEST_DATA` env var for acceptance test and E2E, but actually, they are ignored by acceptance, and E2E have no test about alias that would use postgres yet.
All the other config use it, though.

src/infra/postgres/postgres.go
dolanor: good idea

src/aliases/store/postgres/alias.go
dolanor: I'm ok with it.

src/aliases/store/postgres/alias.go
dolanor: I'm ok with it.

src/aliases/store/postgres/alias.go
dolanor: Done, and I even added the business layer with the entities.

ggarri (approved):

PR #216, opened 2021-07-16
Add radix based policy resolver

nmvalera:
## PR description

Add implementation of a radix tree based privacy resolver.

It covers both
- aggregation procedure of a set of policies
- resolution procedue for a given operation 

## Changelog

- [ ] Add radix based Policy Resolver (https://github.com/consensys/quorum-key-manager/blob/master/CHANGELOG.md).

### DIFF
--- a/src/auth/policy/resolver.go
+++ b/src/auth/policy/resolver.go
@@ -214,7 +214,7 @@ type RadixResolver struct {
 }
 
 // NewRadixResolver creates a new RadixResolver
-func NewRadixResolver(policies ...*types.Policy) (*RadixResolver, error) {
+func NewRadixResolver(policies ...types.Policy) (*RadixResolver, error) {
 	rslvr := &RadixResolver{
 		exactPath:    make(map[string]*actionsResolver),
 		prefixedPath: radix.New(),
@@ -338,9 +338,9 @@ func (r *RadixResolver) insertStatement(sttmnt *statement) error {
 	return nil
 }
 
-func (r *RadixResolver) insertPolicy(policy *types.Policy) error {
+func (r *RadixResolver) insertPolicy(policy types.Policy) error {
 	for _, sttmnt := range policy.Statements {
-		err := r.insertStatement(&statement{*sttmnt, policy})
+		err := r.insertStatement(&statement{*sttmnt, &policy})
 		if err != nil {
 			return err
 		}
--- a/src/auth/policy/resolver.go
+++ b/src/auth/policy/resolver.go
@@ -13,8 +13,8 @@ var (
 )
 
 type Operation struct {
-	Action string
-	Path   string
+	Action       string
+	ResourcePath string
 }
 
 type Result interface {
@@ -46,11 +46,11 @@ func (res *result) Allowed() bool {
 
 func (res *result) Error() error {
 	if res.deny != nil {
-		return fmt.Errorf("action %q on resource %q denied by policy %q statement %q", res.op.Action, res.op.Path, res.deny.policy.Name, res.deny.Name)
+		return fmt.Errorf("action %q on resource %q denied by policy %q statement %q", res.op.Action, res.op.ResourcePath, res.deny.policy.Name, res.deny.Name)
 	}
 
 	if res.allow == nil {
-		return fmt.Errorf("action %q on resource %q not allowed", res.op.Action, res.op.Path)
+		return fmt.Errorf("action %q on resource %q not allowed", res.op.Action, res.op.ResourcePath)
 	}
 
 	return nil
@@ -245,7 +245,7 @@ func (r *RadixResolver) isAuthorized(op *Operation) (res *result) {
 	}
 
 	// do we have an exact match for the path?
-	actionRslvr, hasExact := r.exactPath[op.Path]
+	actionRslvr, hasExact := r.exactPath[op.ResourcePath]
 	if hasExact {
 		// if we have an exact match and it resolves to Deny we return Deny
 		r := actionRslvr.isAuthorized(op)
@@ -256,7 +256,7 @@ func (r *RadixResolver) isAuthorized(op *Operation) (res *result) {
 	}
 
 	// do we have a prefix matching the action?
-	actionPrefix, _, hasPrefix := r.prefixedPath.LongestPrefix(op.Path)
+	actionPrefix, _, hasPrefix := r.prefixedPath.LongestPrefix(op.ResourcePath)
 	if !hasPrefix {
 		// no prefix so we return
 		return
--- a/src/auth/policy/resolver.go
+++ b/src/auth/policy/resolver.go
@@ -56,43 +56,66 @@ func (res *result) Error() error {
 	return nil
 }
 
-type actionResolver struct {
-	exactActions    map[string]*effectsResolver
+// actionsResolver is associated to every statement path
+// it aggregates all statement actions associated to the path into a radix tree
+type actionsResolver struct {
+	// exactActions maps an exact action name to an effect resolver
+	exactActions map[string]*effectResolver
+
+	// prefixedActions maps prefixed action to an effect resolver
 	prefixedActions *radix.Tree
 }
 
-func newActionResolver() *actionResolver {
-	return &actionResolver{
-		exactActions:    make(map[string]*effectsResolver),
+func newActionsResolver() *actionsResolver {
+	return &actionsResolver{
+		exactActions:    make(map[string]*effectResolver),
 		prefixedActions: radix.New(),
 	}
 }
 
-func (rslvr *actionResolver) addStatement(sttmnt *statement) error {
+func (rslvr *actionsResolver) addStatement(sttmnt *statement) error {
+	// loop over all actions in the statement
 	for _, action := range sttmnt.Actions {
-		effect, err := newEffectResolver(sttmnt)
-		if err != nil {
-			return err
-		}
-
 		if strings.HasSuffix(action, wildCardSuffix) {
+			// action is a prefix pattern (e.g. Sign*)
 			actionPrefix := strings.TrimSuffix(action, wildCardSuffix)
-			effects, ok := rslvr.prefixedActions.Get(actionPrefix)
+
+			// lookup for effect already associtated to this prefix
+			effect, ok := rslvr.prefixedActions.Get(actionPrefix)
 			if ok {
-				effects.(*effectsResolver).addEffect(effect)
+				// we already had an effect for prefix so we acculmulate the statement
+				err := effect.(*effectResolver).addStatement(sttmnt)
+				if err != nil {
+					return err
+				}
 			} else {
-				newEffects := new(effectsResolver)
-				newEffects.addEffect(effect)
-				_, _ = rslvr.prefixedActions.Insert(actionPrefix, newEffects)
+				// we create the effect and insert it into the radix tree
+				newEffect := new(effectResolver)
+				err := newEffect.addStatement(sttmnt)
+				if err != nil {
+					return err
+				}
+				_, _ = rslvr.prefixedActions.Insert(actionPrefix, newEffect)
 			}
 		} else {
-			effects, ok := rslvr.exactActions[action]
+			// action is an exact name (e.g. SignTransaction)
+
+			// lookup for effect already associated to this action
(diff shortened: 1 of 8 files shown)


nmvalera:
# Aggregation

Aggregation consists of storing statements with prefix patterns in radix structure so they can be resolved in a timely optimized manner when checking authorizations.

In particular, it avoids having to loop through all statements each time we check authorizations. 

We aggregate into
- one radix tree for resource path prefixes
- for each resource path prefix we have an action tree

Below is an example of aggregation for a 4 statements policy

Radix Policy Resolver

# Resolution

When resolving we visit only the parts of the tree that the operation is concerned about
- if meeting a single Deny the operation is "Denied"
- if not meeting at least once Allow the operation is "Not Allowed"

As soon as a "Deny" is met, we abort the visit of the tree and return a "Deny"

Below are some examples

# Base allowed operation

**Path**: path/bar/test 
**Action**: DoA-Test 

Radix Policy Resolver (1)

In this case

1. We start the visit at path prefix `*` which is a match for `path/bar/test `
2. We visit the corresponding action tree which has no match for `DoA-Test`
3. We continue paths visit up to `path/bar` which is a match
4. We visit the corresponding action tree up to `DoA*` which is a match  for `DoA-Test`
5. `DoA*` match is an `Allow`
6. No other match so we operation is allowed

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
+	if res.allow == nil {
+		return fmt.Errorf("action %q on resource %q not allowed", res.op.Action, res.op.Path)
+	}
+
+	return nil
+}
+
+// actionsResolver is associated to every statement path
+// it aggregates all statement actions associated to the path into a radix tree
+type actionsResolver struct {
```
ggarri, line 61: Why couldn't you use `radix.Tree` to allocate also your `exactAction` all your effectResolver? As I can quickly see from go-radix each node can store a value where we would store the effectResolver, is that right?

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
 	"github.com/consensys/quorum-key-manager/src/auth/types"
 )
 
+var (
+	wildCardSuffix = "*"
+)
+
 type Operation struct {
 	Action string
 	Path   string
```
ggarri, line 17: As far as I can see from our documentation an operation would involve two elements: an action and a resource for instance `key:Get` over `stores::stores/mystore/keys/*`. if so, should we rename `Path` by `ResourcePath` for clarity?

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
+	allow *statement
+	deny  *statement
 }
 
-// Resolver is responsible to control whether an operation is authorized or not
-// depending on the set of policies attached to the resolver
-type Resolver struct {
-	// TODO: to be implemented
+func (res *result) Allowed() bool {
+	return res.deny == nil && res.allow != nil
```
ggarri, line 44: is it possible to obtain this resolt `res.deny == nil and res.allow == nil`? If so we will be denying access even if deny is not added. So that I believe we could simplify this method by:
```
unc (res *result) Allowed() bool {
	return res.allow != nil
```

src/auth/policy/resolver.go
nmvalera: In the radix tree all keys are prefixes. If we also include the exact path into tree then `Sign` would also give access to `SignTransaction` and we don't want that.

src/auth/policy/resolver.go
nmvalera: Updated.

src/auth/policy/resolver.go
nmvalera: Yes this will happen when an operation has no `Deny` and no `Allow`. In which case the operation is not allowed (which is different than denied because there is a `Deny` statement)

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
+	return r.isAuthorized(op)
+}
+
+func (r *RadixResolver) isAuthorized(op *Operation) (res *result) {
+	res = &result{
+		op: op,
+	}
+
+	// do we have an exact match for the path?
+	actionRslvr, hasExact := r.exactPath[op.Path]
```
ggarri: If I would have the following policies, and an user would try to do  `ACTION: keys:Get RESOURCE:stores::stores/mystore/keys/key1`, what would be the result?
```
- effect: Deny
        actions: 
            - keys:Get
        resource:
            - stores::stores/mystore/keys/*
- effect: Allow
        actions: 
            - keys:Get
            - keys:Sign
        resource:
            - stores::stores/mystore/keys/key1
```

Reading from left-to-right I would expect I would expect it to reject, reading right-to-left deny. I personally believe it should be left-to-right, Did you based you decision on how k8s does or any other standard?

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
+	}
+}
+
+// RadixResolver allows to perform authorization checks
+
+// It is built from a set of Policy by aggregating all statements into some
+// radix trees structure allowing optimized authorization checks
+type RadixResolver struct {
+	// exactPath maps an exact path to an action resolver
+	exactPath map[string]*actionsResolver
```
ggarri, line 210: As far as I see we could we don't need `exactPath` and it can use radix.Tree to manage every scenario

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
-func (res *Result) Allowed() bool {
-	return res.isAllowed
+type statement struct {
+	types.Statement
+	policy *types.Policy
 }
 
-func (res *Result) Error() error {
-	return res.err
+type result struct {
```
ggarri, line 36: Why did you need to store all this information as part of the result object. Isn't it enough to return a true or false as the result of resolving the access requested Operation ?

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
 
-// Resolver is responsible to control whether an operation is authorized or not
-// depending on the set of policies attached to the resolver
-type Resolver struct {
-	// TODO: to be implemented
+func (res *result) Allowed() bool {
+	return res.deny == nil && res.allow != nil
+}
+
+func (res *result) Error() error {
```
ggarri, line 47: In which context did you intend to use this errors? This looks more like responsability of the Audit to define which kind of output message it should be added into the auditing logs but for the Store looking to allow or deny and operation this Error() does not seem useful.

src/auth/policy/resolver.go
```diff
@@ -1,41 +1,349 @@
...
+	actionPrefix, _, hasPrefix := rslvr.prefixedActions.LongestPrefix(op.Action)
+	if !hasPrefix {
+		// no prefix so we return
+		return res
+	}
+
+	// we have a prefix
+	// we walk all matching prefix starting from root up to the longest matching prefix
+	// if we meet a single Deny we stop walking and return a Deny
+	rslvr.prefixedActions.WalkPath(actionPrefix, func(_ string, v interface{}) bool {
```
ggarri, line 151: Reading Radix docs (see below) this method appear to navigate top to bottom which implies you will be apply policies `left-to-right`, but using the exactPath you are applying the oppositive behavior of reading `right-to-left` (see comment here https://github.com/ConsenSys/quorum-key-manager/pull/216/files#r671375931) , is there something I am missing here? 

```
// WalkPath is used to walk the tree, but only visiting nodes
// from the root down to a given leaf. Where WalkPrefix walks
// all the entries *under* the given prefix, this walks the
// entries *above* the given prefix.
func (t *Tree) WalkPath(path string, fn WalkFn) {
```

src/auth/policy/resolver.go
ggarri: Then it looks like Radix might not be a good fit since I would expect nodes to be split by a separator char, such as `/` or `:`. I would rather implement my own tree solution if that would be the case

As you describe you would have same issue with resources or action nesting other such as:
```
Action:
- Keys:List
- Keys:ListDeleted

Resources:
- stores::stores/mystoreOne/
- stores::stores/mystore
```

src/auth/policy/resolver.go
nmvalera: In the example above it will "Deny"

src/auth/policy/resolver.go
nmvalera: Not sure to understand.

Radix is used for prefixes (with wildcard last character  `*`) in the case above those are exact path.

src/auth/policy/resolver.go
nmvalera: Hmmm...

First, I am not sure how we could handle cases where you have statements with both the exact path and prefix path (like `/store/keys*` and `/store/keys`).

Second, 
- storing in radix is more expensive in storage than a regular map (because we have intermediary nodes), we accept this extra cost to accelerate prefix pattern resolution
- for an exact path, a normal map is less expensive in storage and enables faster queries than a radix

So having both a radix for prefixes and a map for exact sounds the best option (btw this is also how it is implemented in Vault).

src/auth/policy/resolver.go
nmvalera: To return extra information details about why an operation was rejected to a user. It helps users configure/debug their policies.

Typically when having a large number of policies, it would be almost impossible to properly configure/debug the policies.

Btw this is what AWS does.

darioAnongba:
Hi! I started reviewing the Policy resolver. I'm probably not understanding this feature and I'm sorry to voice concerns now that it's already implemented but I already made this remark before. I consider this implementation to be completely over-engineered. Maybe I'm missing something but:
- Why are these policies needed in the first place ?
- What are the use cases where these are needed
- Why did we decide to go for such an implementation if (as far as I know) the different Vault solutions do not implement such fine-grained rules.

From my point of view this is not needed because the QKM only deals with creation and usage of keys and secrets:
- A secret can either be read or written. It makes no sense to allow someone to create a secret but not update it. Actions are only READ, WRITE.
- A key can either be used to sign or to encrypt. We either do the authorization per user with READ, CREATE, SIGN, ENCRYPT or on key creation (like AKV does) with SIGN, ENCRYPT

My point is that fine-grained authorization is not needed and will never be. It makes no sense to allow someone to SignTransaction but not SignEEATransaction. Now that we have a DB, we can store who is the user creating the resource and we can also store which operations are supported on a given resource, dropping this whole implementation entirely right?

cc: @ggarri  , @julien-marchand , @nmvalera

darioAnongba:
Also setting policies via an external file sounds weird to me. At some point we will be able to create stores and nodes via API and we will store them in DB. Manifests will be deprecated. If so, setting of policies over created resouces should be done per resource or per user (with custom claims).

Again, I'm probably missing something here

src/auth/policy/resolver.go
ggarri: But that is exactly what I am trying to avoid, using two independent flows for same feature. To mitigate this issue you describe above ` Sign would be treated as a prefix and would also give access to SignTransaction`, couldn't you just store as part of the node data (`effectResolver`) whether or not the path is a exactPath or a wildcarded one? so at the time you resolve you can deal with the issue of `Sign` vs `SignTransaction`

src/auth/policy/resolver.go
ggarri: That is right, I would also expect deny due to your first statement, but in the method `isAuthorized(op)` in that particular operation you will fall into the `exactPath` branch of your code:
```
func (r *RadixResolver) isAuthorized(op *Operation) (res *result) {
	res = &result{
		op: op,
	}

	// do we have an exact match for the path?
	actionRslvr, hasExact := r.exactPath[op.Path]
	if hasExact {
		// if we have an exact match and it resolves to Deny we return Deny
		r := actionRslvr.isAuthorized(op)
		if r.deny != nil {
			return r
		}
		res = r
	}
```

what would make it `allowed` if I can read this code propertly

src/auth/policy/resolver.go
ggarri: In terms of performance we are comparing a log(n) with O1, what IMO does not pay off the complexity of having two separate flows to maintain in the code and the additional complexity that means. I cannot see strong reasons to not use just radix for both alternatives, it might be worthy to have a chat about it to explore how it could be done.

PR #217, opened 2021-07-16
Feature/115 auth apikey

6l20:
## PR description

Add API-KEYS auth management

## Changelog

- Add API-KEY authenticator
- Add API-KEYS app parameter
- Parses API-KEYS file
- Compare Base64 encoded API-KEY with stored hash
- Accepts identical
- Rejects different or badly encoded ones
- Ignores when missing


ggarri (changes requested):
Would you mind to decouple the tls changes here and only include things related to api-key?

cmd/flags/auth.go
```diff
@@ -178,3 +211,42 @@ func issuerCertificates(vipr *viper.Viper) ([]*x509.Certificate, error) {
...
+		return nil, err
+	}
+	defer func(csvfile *os.File) {
+		_ = csvfile.Close()
+	}(csvfile)
+
+	// Parse the file
+	r := csv.NewReader(csvfile)
+	// Set separator as ";"
+	r.Comma = ';'
```
ggarri, line 229: could you make separator char and comment chat a constant so it is easier to locate for future changes?

ggarri, line 229: could you make separator char and comment chat a constant so it is easier to locate for future changes?

cmd/flags/auth.go
```diff
@@ -178,3 +211,42 @@ func issuerCertificates(vipr *viper.Viper) ([]*x509.Certificate, error) {
...
+		// Read each line from csv
+		cells, err := r.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		retFile[cells[0]] = &apikey.UserNameAndGroups{UserName: cells[1],
```
ggarri, line 246: I am missing the entire error handling here, we would panic in case csv is not formatted as expected, instead we should manage that and return to the user a proper error message

pkg/tls/testutils/testutils.go
```diff
@@ -107,4 +107,40 @@ ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS
 R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp
 -----END CERTIFICATE-----
 `
+	TLSAuthKey = `-----BEGIN EC PARAMETERS-----
```
ggarri, line 110: is this part of this PR?

src/auth/authenticator/api-key/authenticator.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Authenticate checks APIKEY hashes retrieve user Info
+// ? -> Username
+// ? -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract ApiKey
+	clientAPIKey, ok, err := extractAPIKey(req.Header.Get("Authorization"), authenticator.B64Encoder)
+	if err != nil {
+		// could not be decoded
+		return nil, errors.UnauthorizedError("apikey format error")
```
ggarri, line 46: error messages must be describe and never reused. Di you think you could provide a more clear and unique message here and in the following error?

src/auth/authenticator/api-key/authenticator.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	clientAPIKeyHash := h.Sum(nil)
+
+	// search hex string hashes
+	strClientHash := hex.EncodeToString(clientAPIKeyHash)
+	// Upper case hash
+	strClientHash = strings.ToUpper(strClientHash)
```
ggarri, line 65: I assume you are trying to mitigate a case sensitive issue here, do you think this is needed? IMO a not exact match on your key should lead to an authorization error

src/auth/authenticator/api-key/authenticator.go
```diff
@@ -0,0 +1,88 @@
...
+	userAndGroups, contains := authenticator.APIKeyFile[strClientHash]
+	if contains {
```
ggarri, line 67: This could be simplified by:
```
	if userAndGroups, ok := authenticator.APIKeyFile[strClientHash]; ok {
```

src/auth/authenticator/api-key/authenticator.go
```diff
@@ -0,0 +1,88 @@
...
+// ? -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract ApiKey
+	clientAPIKey, ok, err := extractAPIKey(req.Header.Get("Authorization"), authenticator.B64Encoder)
+	if err != nil {
+		// could not be decoded
+		return nil, errors.UnauthorizedError("apikey format error")
+	}
+	// respond nothing when none found
+	if !ok || clientAPIKey == "" {
```
ggarri, line 49: As far as I can see `ok == true && clientAPIKey == ""` cannot happen without an error calling `extractAPIKey()`

src/auth/authenticator/api-key/config.go
```diff
@@ -0,0 +1,25 @@
+package apikey
+
+import (
+	"encoding/base64"
+	"hash"
+)
+
+type Config struct {
+	APIKeyFile map[string]*UserNameAndGroups
```
ggarri, line 9: why are you using pointer to that struct instead of the struct itself?

cmd/flags/auth.go
6l20: Ok

pkg/tls/testutils/testutils.go
6l20: I thought I had all the TLS parts removed... will check

cmd/flags/auth.go
6l20: Yes that part is missing, I will had it

6l20: I would rather return a nil config instead of panicing

src/auth/authenticator/api-key/authenticator.go
6l20: I will push up the underlying error here

src/auth/authenticator/api-key/authenticator.go
6l20: well it seems that some SHA functions use upper hex digits some don't. The idea is to force upper case

src/auth/authenticator/api-key/authenticator.go
6l20: Yes, right. I did simplify this bit

src/auth/authenticator/api-key/config.go
6l20: I don't know. I tend to mostly use pointers to structs in maps. UserNameAndGroups is not that big, a habit of mine I would say :-)

src/auth/authenticator/api-key/authenticator.go
6l20: Right

src/auth/authenticator/api-key/config.go
ggarri: but you won't safe any memory or anything since the map is filled only in case elements a fullfilled, not the same than having struct with properties which can be empty where in that case you woud safe space

ggarri (approved):

src/auth/authenticator/api-key/authenticator.go
6l20: I removed the ToUpper thing

src/auth/authenticator/api-key/config.go
6l20: Well actually I suppose it would save some space when building the Authenticator in case we have a huge csv file with many keys and long rows. A map of pointers has a smaller mem foot print than a map of structs.

PR #218, opened 2021-07-20
Fix/qkm 215

### LINKED ISSUE #215: Fix error handling
Since the implementation of Connectors, the error handling is wrong because we do not log where the error happens. This means that we return an internal error to the users instead of logging it and mapping it to a suitable error message. All stores are affected.

darioAnongba:
## Fixed Issue(s)
fixes #215 by rewriting the error handling + some bugs in the connectors

### DIFF
--- a/src/stores/store/keys/keys.go
+++ b/src/stores/store/keys/keys.go
@@ -70,7 +70,7 @@ func VerifySignature(logger log.Logger, pubKey, data, sig []byte, algo *entities
 	case algo.EllipticCurve == entities.Bn254 && algo.Type == entities.Eddsa:
 		verified, err = crypto.VerifyEDDSASignature(pubKey, data, sig)
 	default:
-		errMessage := "unsupported algorithm"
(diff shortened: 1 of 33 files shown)


src/stores/store/keys/keys.go
```diff
@@ -70,7 +70,7 @@ func VerifySignature(logger log.Logger, pubKey, data, sig []byte, algo *entities
 	case algo.EllipticCurve == entities.Bn254 && algo.Type == entities.Eddsa:
 		verified, err = crypto.VerifyEDDSASignature(pubKey, data, sig)
 	default:
-		errMessage := "unsupported algorithm"
+		errMessage := "unsupported signing algorithm and "
```
6l20, line 73: Is end of message missing ?

6l20 (approved):
One minor comment.

src/stores/store/keys/keys.go
darioAnongba: good catch, fixed.

Issue #196, opened 2021-07-08
Replace the in-memory by DB for eth1store


ggarri:
## Acceptance test
- [ ] Replace the in-memory by DB to store eth1 accounts
- [ ] Skip sync-up for eth1 account on QKM boot up and replace it by a CLI command

PR #224, opened 2021-07-27
Implement Postgres DB for ETH1 store

darioAnongba:
## PR description

- Implements Eth1 accounts using Postgres DB. this implementation could be better if all key stores used a DB as well instead of having some key stores using it and others not so we could create links in the DB with a HAS ONE relationship.
- For the moment the ETH1Accounts model is independent and does not contain a Key.
- Removes inconsistent fields like `expireAt` and `destroyedAt` from the ETH1Accounts model.

## Fixed Issue(s)
Fixey #196


src/stores/api/formatters/eth1.go
```diff
@@ -88,18 +88,15 @@ func FormatEEATransaction(tx *types.SignEEATransactionRequest) (*ethtypes.Transa
...
 	return &types.Eth1AccountResponse{
-		KeyID:               key.ID,
-		Address:             key.Address,
-		PublicKey:           key.PublicKey,
-		CompressedPublicKey: key.CompressedPublicKey,
-		Tags:                key.Tags,
-		Disabled:            key.Metadata.Disabled,
-		CreatedAt:           key.Metadata.CreatedAt,
-		UpdatedAt:           key.Metadata.UpdatedAt,
-		ExpireAt:            key.Metadata.ExpireAt,
```
ggarri, line 101: as far as I understand it if a key has an expiration time the account linked to it should have the same value and therefore users should be informed about it

src/stores/manager/local/eth1_account.go
```diff
@@ -61,57 +59,5 @@ func NewEth1(ctx context.Context, specs *Eth1Specs, eth1Accounts database.ETH1Ac
...
 
-	err = InitDB(ctx, keyStore, eth1Accounts)
-	if err != nil {
-		return nil, err
-	}
-
-	return eth1.New(keyStore, eth1Accounts, logger), nil
-}
-
-func InitDB(ctx context.Context, keyStore keys.Store, db database.ETH1Accounts) error {
```
ggarri, line 72: IMO this code could be moved into a CLI cmd in oder to perform a DB synchonization on demand bases

ggarri (approved):

src/stores/manager/local/eth1_account.go
darioAnongba: done: https://app.zenhub.com/workspaces/orchestrate-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/227

src/stores/api/formatters/eth1.go
darioAnongba: `expireAt` is not implemented at the moment, we only have the concept of deleting and destroying. Deleting is equivalent to soft-delete and destroying completely destroys the keys. We do not correlate the behavior of the underlying vaults (like AWS that deletes after 30 days) in the QKM, thefore expireAt is not needed.

Issue #231, opened 2021-07-28
Use a DB for all stores


darioAnongba:
**Main reason of the change** (bug):
- The resources of the QKM live inside stores, that are sets of secrets, keys and eth1 accounts. For us to be able to separate these resources and deny access to those that do not have it, we must use a DB to store these entities and link them with their respective store.

_Current behavior_
One instance of the QKM (SaaS) with 2 eth1 stores `A` and `B`. Currently, if `A` contains an account, `B` will be able to get it because we do not filter select operations by storeID (because we don't store resources in DB at all or the resources in DB are not filtered by storeID).

_Expected behavior_
One instance of the QKM with 2 eth1 stores `A` and `B`. `B` cannot access resources of `A` and vice-versa. note that has nothing to do with multi-tenancy.

**Other reasons for the change** (features):
- We are planning to use a DB for AWS and Hashicorp key stores because the underlying vaults do not support some behaviors that we would like to force in all stores, such as deleting before destroying.
- Currently, some stores such as the eth1 store and the local key store use a DB to store eth1 accounts and keys. This is inconsistent compared to the other stores such as AKV and Hashicorp stores.
- When you create a store, it should be initially empty. Currently, some stores will be and others will automatically import all resources

**Acceptance criteria**
- Move the logic of the usage of the DB and the DB transactions to the connectors. Modify the interface of the connectors to accept a storeID param (note that the stores interface and the connectors interface will no longer be the same).

PR #240, opened 2021-07-29
Use a DB for all key stores

darioAnongba:
## PR description
This PR moves the business logic to the connector and uses the stores as data layers to write and read data from the underlying Vaults.
The connectors now become the list of use cases of our domains (here keys only) and orchestrate the flow between logger, vault and DB. The interface of connectors is not the same as the store interface anymore.

## Fixed Issue(s)
Part of #231 

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://github.com/consensys/quorum-key-manager/blob/master/CHANGELOG.md).


src/stores/connectors/keys/create.go
```diff
@@ -0,0 +1,24 @@
...
+	logger.Debug("creating key")
+
+	key, err := c.store.Create(ctx, id, alg, attr)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.db.Keys().Add(ctx, key)
+	if err != nil {
+		return nil, err
```
6l20, line 19: Shouldn't we "UnCreate" the created key in that case ?

6l20 (approved):
I left a few comments as questions and further discussions points.

src/stores/connectors/keys/delete.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func (c Connector) Delete(ctx context.Context, id string) error {
+	logger := c.logger.With("id", id)
+	logger.Debug("deleting key")
+
+	err := c.db.RunInTransaction(ctx, func(dbtx database.Database) error {
+		err := dbtx.Keys().Delete(ctx, id)
+		if err != nil {
+			return err
```
ggarri, line 15: This error will be handled to the users, we should be logging here and rephasing the message to something like "failed to persist key"

src/stores/connectors/keys/list.go
```diff
@@ -0,0 +1,33 @@
+package keys
+
+import "context"
+
+func (c Connector) List(ctx context.Context) ([]string, error) {
+	ids := []string{}
+	keysRetrieved, err := c.db.Keys().GetAll(ctx)
```
ggarri, line 7: Alternatively we could read IDs directly from DB for optimization proposes

src/stores/store/models/testutils/faker.go
```diff
@@ -1,9 +1,9 @@
 package testutils
 
 import (
+	"github.com/consensys/quorum-key-manager/src/stores/store/models"
 	"time"
 
-	"github.com/consensys/quorum-key-manager/src/stores/store/database/models"
```
ggarri, line 6: Why did you move your models one level up, as far as I see it models do not make sense out of database/, same as migrations

src/stores/connectors/keys/get.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"github.com/consensys/quorum-key-manager/src/stores/store/entities"
+)
+
+func (c Connector) Get(ctx context.Context, id string) (*entities.Key, error) {
+	logger := c.logger.With("id", id)
+
+	key, err := c.db.Keys().Get(ctx, id)
```
ggarri, line 11: is this query excluding soft_deleted items ?

src/stores/connectors/keys/list.go
ggarri: is this query excluding soft_deleted items ?

src/stores/connectors/keys/update.go
```diff
@@ -0,0 +1,38 @@
...
+	logger.Debug("updating key")
+
+	key, err := c.db.Keys().Get(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	key.Tags = attr.Tags
+
+	err = c.db.RunInTransaction(ctx, func(dbtx database.Database) error {
+		derr := c.db.Keys().Update(ctx, key)
```
ggarri, line 20: is go_pg automatically updating the field `updated_at` or should you take care of it?

src/stores/store/eth1/local/parser.go
```diff
@@ -1,12 +1,12 @@
...
 
 import (
-	"github.com/consensys/quorum-key-manager/src/stores/store/database/models"
 	"github.com/consensys/quorum-key-manager/src/stores/store/entities"
+	"github.com/consensys/quorum-key-manager/src/stores/store/models"
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-func parseKey(key *entities.Key, attr *entities.Attributes) *models.ETH1Account {
+func parseKey(key *models.Key, attr *entities.Attributes) *models.ETH1Account {
```
ggarri, line 9: Trying to apply a more strict idea of how Hex Arch work, models should never live outside of our /database, our entities are the respentation of our application logic and those are the elements we should move around. Therefore I would suggest to encapsulate all this under /database aiming to prevent same issues we go in Orchestrate where we got recurrent questions about when we should use models o entities.

src/stores/store/keys/akv/akv.go
```diff
@@ -30,11 +28,7 @@ func New(client akv.KeysClient, logger log.Logger) *Store {
 	}
 }
 
-func (s *Store) Info(context.Context) (*entities.StoreInfo, error) {
-	return nil, errors.ErrNotImplemented
-}
-
-func (s *Store) Create(ctx context.Context, id string, alg *entities.Algorithm, attr *entities.Attributes) (*entities.Key, error) {
+func (s *Store) Create(ctx context.Context, id string, alg *entities.Algorithm, attr *entities.Attributes) (*models.Key, error) {
```
ggarri, line 31: IMO stores should always work with our entities, and models are a external reprensetation of our storage which never should leak into our application to reduce the potential impact of changes over our DB model

src/stores/connectors/keys/create.go
```diff
@@ -0,0 +1,24 @@
+package keys
```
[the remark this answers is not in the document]
darioAnongba: I prefer one file per use case than one big file that implements every use case and exceeds the 300 lines.

src/stores/connectors/keys/import.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+	"github.com/consensys/quorum-key-manager/src/stores/store/entities"
+)
+
+func (c Connector) Import(ctx context.Context, id string, privKey []byte, alg *entities.Algorithm, attr *entities.Attributes) (*entities.Key, error) {
+	logger := c.logger.With("id", id).With("algorithm", alg.Type).With("curve", alg.EllipticCurve)
+	logger.Debug("importing key")
+
+	key, err := c.store.Import(ctx, id, privKey, alg, attr)
```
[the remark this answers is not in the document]
darioAnongba: how would you implement this otherwise? We need a mechanism to encrypt and decrypt private keys when you want to import them? Every wallet out there (Ledger, Metamask, etc.) allows you to import a "Seed", which is your private key, the difference is that these wallets are local

src/stores/connectors/keys/verify.go
```diff
@@ -0,0 +1,39 @@
+package keys
+
+import (
+	"context"
+	"github.com/consensys/quorum-key-manager/pkg/crypto"
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/consensys/quorum-key-manager/src/stores/store/entities"
+)
+
+func (c Connector) Verify(_ context.Context, pubKey, data, sig []byte, algo *entities.Algorithm) error {
```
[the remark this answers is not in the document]
darioAnongba: no, pubKey is a parameter of the function. You are not verifying that a signature belongs to a key in your DB, you are verifying any signature belongs to any pubKey. This function should not even be under "stores" at all.

src/stores/store/eth1/local/parser.go
dolanor: It seems again we have a different naming of things.
I thought that entities were directly linked to database rows; and models more connected to the business/domain.

src/stores/connectors/keys/list.go
darioAnongba: won't have time but I created a ticket: https://app.zenhub.com/workspaces/orchestrate-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/241

src/stores/store/models/testutils/faker.go
darioAnongba: reverted

src/stores/connectors/keys/get.go
darioAnongba: adding tests

src/stores/connectors/keys/update.go
darioAnongba: adding tests

src/stores/store/eth1/local/parser.go
darioAnongba: moved to using models

src/stores/store/keys/akv/akv.go
darioAnongba: moved to using models

src/stores/connectors/keys/import.go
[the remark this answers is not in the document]
darioAnongba: letting this ticket to a different topic

src/stores/connectors/keys/create.go
darioAnongba: to discuss

darioAnongba: won't have time to discuss this so I created a ticket: https://app.zenhub.com/workspaces/orchestrate-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/242

src/stores/connectors/keys/get.go
darioAnongba: tests added

src/stores/connectors/keys/update.go
darioAnongba: tests added

Issue #219, opened 2021-07-20
Re-write the Ethereum sign endpoint


darioAnongba:
Signing a message using an Ethereum account is defined by an [EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md) and is different than signing a transaction. The current implementation of the signing endpoint is not EIP-191 compliant.

**References**:
- EIP-191: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md
- Web3JS definition: https://web3js.readthedocs.io/en/v1.3.4/web3-eth-accounts.html#sign
- EthersJS definition: https://docs.ethers.io/v5/api/signer/#Signer--signing-methods

PR #243, opened 2021-08-02
Feature/219 eth message sign

6l20:
## PR description

Implementation of the EIP-191 signature scheme.

## Fixed Issue(s)
#219 

## Changelog


src/stores/api/handlers/eth1.go
```diff
@@ -43,6 +43,7 @@ func (h *Eth1Handler) Register(r *mux.Router) {
 	r.Methods(http.MethodPost).Path("/{address}/sign-quorum-private-transaction").HandlerFunc(h.signPrivateTransaction)
 	r.Methods(http.MethodPost).Path("/{address}/sign-eea-transaction").HandlerFunc(h.signEEATransaction)
 	r.Methods(http.MethodPost).Path("/{address}/sign-typed-data").HandlerFunc(h.signTypedData)
+	r.Methods(http.MethodPost).Path("/{address}/sign-eip191-data").HandlerFunc(h.signEIP191Data)
```
ggarri, line 46: it should be called `sign-data` only and the old `sign-data` rename to `sign-hash`

src/stores/store/eth1/eth1.go
```diff
@@ -71,7 +74,7 @@ type Store interface {
 	SignPrivate(ctx context.Context, addr string, tx *quorumtypes.Transaction) ([]byte, error)
 
 	// ECRevocer returns the Ethereum address from a signature and data
-	ECRevocer(ctx context.Context, data, sig []byte) (string, error)
+	ECRecover(ctx context.Context, data, sig []byte) (string, error)
```
nmvalera, line 77: You can also update docstring. Thx

src/stores/api/types/eth1.go
```diff
@@ -28,6 +28,10 @@ type SignHexPayloadRequest struct {
 	Data hexutil.Bytes `json:"data" validate:"required" example:"0xfeee" swaggertype:"string"`
 }
 
+type SignMessagePayloadRequest struct {
+	Data hexutil.Bytes `json:"message" validate:"required" example:"any message" swaggertype:"string"`
```
ggarri, line 32: if value is an hexutils.Bytes, example cannot be "any message"

src/stores/store/eth1/local/local.go
```diff
@@ -406,6 +412,13 @@ func getEIP712EncodedData(typedData *core.TypedData) (string, error) {
 	return fmt.Sprintf("\x19\x01%s%s", domainSeparatorHash, typedDataHash), nil
 }
 
+// getEIP191EncodedData encodes the given message that can be later recovered
+// with the given validator. following EIP-171 spec version 0x0
+func getEIP191EncodedData(address string, msg []byte) string {
```
nmvalera, line 417: MINOR: feels like `getEIP191EncodedData` could return bytes so it avoids to the transformation in the caller function.

src/stores/connectors/eth1.go
```diff
@@ -192,6 +192,20 @@ func (c Eth1Connector) SignData(ctx context.Context, addr string, data []byte) (
...
+
+func (c Eth1Connector) SignEIP191Data(ctx context.Context, addr string, data []byte) ([]byte, error) {
+	logger := c.logger.With("address", addr)
+
+	result, err := c.store.SignEIP191Data(ctx, addr, data)
+	if err != nil {
+		c.logger.WithError(err).Error("failed to sign data")
+		return nil, err
+	}
+
```
nmvalera, line 208: Log message below could be updated to indicate `EIP191`

src/stores/connectors/eth1.go
```diff
@@ -192,6 +192,20 @@ func (c Eth1Connector) SignData(ctx context.Context, addr string, data []byte) (
...
+	return result, nil
+
+}
+
+func (c Eth1Connector) SignEIP191Data(ctx context.Context, addr string, data []byte) ([]byte, error) {
+	logger := c.logger.With("address", addr)
+
+	result, err := c.store.SignEIP191Data(ctx, addr, data)
+	if err != nil {
+		c.logger.WithError(err).Error("failed to sign data")
```
nmvalera, line 205: Log message below could be updated to indicate `EIP191`

src/stores/store/eth1/local/local.go
ggarri, line 417: IMO if this function is called only once and it contains only a single line of code there is not real value in doing this encapulation aside so I would rather place the code itself in `SignEIP191Data`

src/stores/api/handlers/eth1.go
```diff
@@ -275,6 +276,46 @@ func (h *Eth1Handler) signData(rw http.ResponseWriter, request *http.Request) {
...
+		return
+	}
+
+	signature, err := eth1Store.SignEIP191Data(ctx, getAddress(request), signPayloadReq.Data)
+	if err != nil {
+		WriteHTTPErrorResponse(rw, err)
+		return
+	}
+
+	_, _ = rw.Write([]byte(hexutil.Encode(signature)))
```
ggarri, line 316: The format of the response must matched this one https://web3js.readthedocs.io/en/v1.3.4/web3-eth-accounts.html#id14

src/stores/store/eth1/local/local.go
```diff
@@ -406,6 +412,13 @@ func getEIP712EncodedData(typedData *core.TypedData) (string, error) {
...
+	return fmt.Sprintf("\x19\x00%s%s", address, string(msg))
```
ggarri, line 418: Why did you choose to concatenate data to sign that way?
```
Version 0 has <20 byte address> for the version specific data, and the address is the intended validator. In the case of a Multisig wallet, that is the wallet's own address .
```

instead of following this one?
```
Additionally, 0x19 has been chosen because since ethereum/go-ethereum#2940 , the following is prepended before hashing in personal_sign:
```

In addition I would rather work with byes all along to avoid enconding issues

6l20: I did choose to keep that format to be consistent with the already existing getEIP712EncodedData. But I agree with you bytes would be better. Should we change getEIP712EncodedData too then ?

src/stores/api/handlers/eth1.go
6l20: Ok

src/stores/store/eth1/eth1.go
6l20: Ok

src/stores/store/eth1/local/local.go
6l20: Yes right, but I chose that format to be consistent with existing `getEIP712EncodedData`

6l20: I would say it helps reading the EIP-191 transformation more clearly.
I don't like the get name that much, just kept because of the existing EIP-172 thing already existing

6l20: Right, I will change to []byte

src/stores/store/eth1/local/local.go
```diff
@@ -406,6 +418,16 @@ func getEIP712EncodedData(typedData *core.TypedData) (string, error) {
...
 }
 
+// getEIP191EncodedData encodes the given message that can be later recovered
+// with the given validator. following EIP-191 spec version 0x0
+func getEIP191EncodedData(address string, msg []byte) (eipFormat, msgHash []byte, err error) {
+	hexAddress, err := hex.DecodeString(address)
+	if err != nil {
+		return nil, nil, err
+	}
+	return append([]byte{0x19, 0x00}, append(hexAddress, msg...)...), crypto.Keccak256(msg), nil
```
ggarri, line 428: As far as I understand we are using a `personal_sign` so we should be appending 0x45 instead of 0x00

image

src/stores/store/eth1/local/local.go
```diff
@@ -406,6 +418,16 @@ func getEIP712EncodedData(typedData *core.TypedData) (string, error) {
 	return fmt.Sprintf("\x19\x01%s%s", domainSeparatorHash, typedDataHash), nil
 }
 
+// getEIP191EncodedData encodes the given message that can be later recovered
+// with the given validator. following EIP-191 spec version 0x0
+func getEIP191EncodedData(address string, msg []byte) (eipFormat, msgHash []byte, err error) {
```
ggarri, line 423: Having `msgHash` as part of the response of this method seems to not have any value added, do you think you could remove it?

src/stores/store/eth1/local/local.go
```diff
@@ -216,6 +217,17 @@ func (s *Store) SignTypedData(ctx context.Context, addr string, typedData *core.
 	return s.Sign(ctx, addr, []byte(encodedData))
 }
 
+func (s *Store) SignEIP191Data(ctx context.Context, addr string, data []byte) (sig, msgHash []byte, err error) {
```
ggarri, line 220: `msgHash` is only a keckac(data) of the input data. I would suggest to simplify this interface to only return with `sig` and `error`

src/stores/api/types/eth1.go
```diff
@@ -28,6 +28,19 @@ type SignHexPayloadRequest struct {
 	Data hexutil.Bytes `json:"data" validate:"required" example:"0xfeee" swaggertype:"string"`
 }
 
+type SignMessagePayloadRequest struct {
+	Data hexutil.Bytes `json:"message" validate:"required" example:"0xabcd1234" swaggertype:"string"`
+}
+
+type SignMessagePayloadResponse struct {
+	Message     string `json:"message" example:"any message goes here" swaggertype:"string"`
```
ggarri, line 36: Message example is incorrect since we expect message to be passed as hex value

6l20: As I understand it the message should be a human readable message

ggarri: But the `message` will correspond to what the user provides via API therefore it will be hex value:
```
type SignMessagePayloadRequest struct {
	Data hexutil.Bytes `json:"message" validate:"required" example:"0xabcd1234" swaggertype:"string"`
```

ggarri (approved):

src/stores/store/eth1/local/local.go
ggarri: After doing another research I think you are right and we could use `0x00`

src/stores/store/eth1/eth1.go
```diff
@@ -56,13 +56,14 @@ type Store interface {
 	Sign(ctx context.Context, addr string, data []byte) ([]byte, error)
 
 	// SignData signs a hash using the specified Ethereum account
-	SignData(ctx context.Context, addr string, data []byte) ([]byte, error)
+	SignData(ctx context.Context, addr string, data []byte) (sig, msg, msgHash []byte, err error)
```
ggarri, line 59: This should be called SignHash to match API naming

src/stores/store/eth1/eth1.go
```diff
@@ -56,13 +56,14 @@ type Store interface {
...
 
 	// SignTypedData signs EIP-712 formatted data using the specified Ethereum account
 	SignTypedData(ctx context.Context, addr string, typedData *core.TypedData) ([]byte, error)
 
 	// SignEIP191Data signs EIP-191 formatted data using the specified Ethereum account
-	SignEIP191Data(ctx context.Context, addr string, data []byte) (sig []byte, msgHash []byte, err error)
+	// returns the EIP the signature, formatted message and the msgHash
+	SignEIP191Data(ctx context.Context, addr string, data []byte) (sig, msg, msgHash []byte, err error)
```
ggarri, line 66: I would rather follow naming of api an call this SignData

src/stores/connectors/eth1.go
```diff
@@ -184,30 +184,30 @@ func (c Eth1Connector) Sign(ctx context.Context, addr string, data []byte) ([]by
 	return result, nil
 }
 
-func (c Eth1Connector) SignData(ctx context.Context, addr string, data []byte) ([]byte, error) {
+func (c Eth1Connector) SignData(ctx context.Context, addr string, data []byte) (sig, msg, msgHash []byte, err error) {
```
ggarri, line 187: You added two returned values but they are always empty

src/stores/store/eth1/local/local.go
```diff
@@ -216,10 +217,18 @@ func (s *Store) SignTypedData(ctx context.Context, addr string, typedData *core.
 	return s.Sign(ctx, addr, []byte(encodedData))
 }
 
+func (s *Store) SignEIP191Data(ctx context.Context, addr string, data []byte) (sig, msg, msgHash []byte, err error) {
```
ggarri, line 220: As we discussed it would be better to have 1:1 naming correspondency with the API, therefore it should be called SignData

ggarri (approved):

src/stores/api/handlers/eth1.go
```diff
@@ -38,11 +38,12 @@ func (h *Eth1Handler) Register(r *mux.Router) {
 	r.Methods(http.MethodPost).Path("").HandlerFunc(h.create)
 	r.Methods(http.MethodPost).Path("/import").HandlerFunc(h.importAccount)
 	r.Methods(http.MethodPost).Path("/{address}/sign").HandlerFunc(h.sign)
-	r.Methods(http.MethodPost).Path("/{address}/sign-data").HandlerFunc(h.signData)
+	r.Methods(http.MethodPost).Path("/{address}/sign-hash").HandlerFunc(h.signData)
```
ggarri, line 41: Remeber to update the name of the func to `signHash`

src/stores/store/eth1/local/local.go
6l20: This is the first place where we are able to compute it, I thought it would be good to do so here.

src/stores/api/types/eth1.go
6l20: Yes, thats true if we follow the input spec but I was wondering if that input :
type SignMessagePayloadRequest struct {
	Data string `json:"message" validate:"required" example:"any text message" swaggertype:"string"`
}
would not be more appropriate 🤔 . I really cannot see the point of signing a non readable message with that endpoint.

src/stores/store/eth1/eth1.go
6l20: Yes agreed

src/nodes/interceptor/eth_sign.go
```diff
@@ -19,7 +19,7 @@ func (i *Interceptor) ethSign(ctx context.Context, from ethcommon.Address, data
 		return nil, err
 	}
 
-	sig, err := store.Sign(ctx, from, data)
+	sig, err := store.SignMessage(ctx, from, data.String())
```
ggarri, line 22: This is intended to use same behaviour than geth does when user calls `eth_sign`. Looking at `go-ethereum` code base it was doing raw signing  using cryptp.ecdsa as you can see in the following code:

https://github.com/ethereum/go-ethereum/blob/295bc35ecf9b29f72db8d71f61e2d65b4d4bb24c/signer/core/signed_data.go#L129

ggarri: We might need to keep the Sign method as part of our store BUT we don't expose it in our API

src/stores/api/handlers/eth1.go
```diff
@@ -37,15 +37,15 @@ func NewAccountsHandler(s stores.Manager) *Eth1Handler {
...
-	r.Methods(http.MethodPost).Path("/{address}/sign").HandlerFunc(h.sign)
-	r.Methods(http.MethodPost).Path("/{address}/sign-hash").HandlerFunc(h.signHash)
 	r.Methods(http.MethodPost).Path("/{address}/sign-transaction").HandlerFunc(h.signTransaction)
 	r.Methods(http.MethodPost).Path("/{address}/sign-quorum-private-transaction").HandlerFunc(h.signPrivateTransaction)
 	r.Methods(http.MethodPost).Path("/{address}/sign-eea-transaction").HandlerFunc(h.signEEATransaction)
 	r.Methods(http.MethodPost).Path("/{address}/sign-typed-data").HandlerFunc(h.signTypedData)
+	r.Methods(http.MethodPost).Path("/{address}/sign-message").HandlerFunc(h.signMessage)
 	r.Methods(http.MethodPut).Path("/{address}/restore").HandlerFunc(h.restore)
 	r.Methods(http.MethodPost).Path("/ec-recover").HandlerFunc(h.ecRecover)
 	r.Methods(http.MethodPost).Path("/verify-signature").HandlerFunc(h.verifySignature)
```
ggarri, line 47: Minor: I would rather remove the suffix 'signature' from our endpoints since it looks very redundant

src/stores/api/types/testutils/faker.go
```diff
@@ -63,9 +63,9 @@ func FakeUpdateEth1AccountRequest() *types.UpdateEth1AccountRequest {
 	}
 }
 
-func FakeSignHexPayloadRequest() *types.SignHexPayloadRequest {
-	return &types.SignHexPayloadRequest{
-		Data: hexutil.MustDecode("0xfeee"),
+func FakeSignMessageRequest() *types.SignMessageRequest {
+	return &types.SignMessageRequest{
+		Message: []byte("any message goes here"),
```
ggarri, line 68: isnt it required an hex value as Message?

src/stores/connectors/eth1/utils.go
```diff
@@ -0,0 +1,26 @@
...
+
+	domainSeparatorHash, err := typedData.HashStruct(formatters.EIP712DomainLabel, typedData.Domain.Map())
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("\x19\x01%s%s", domainSeparatorHash, typedDataHash), nil
+}
+
+func getEIP191EncodedData(msg string) string {
```
ggarri, line 24: I thought we decided on working only with bytes when we are on our internal app flows

tests/acceptance/store_test_eth1.go
```diff
@@ -178,9 +178,9 @@ func (s *eth1TestSuite) TestList() {
 	})
 }
 
-func (s *eth1TestSuite) TestSignVerify() {
+func (s *eth1TestSuite) TestSignMessageVerify() {
 	ctx := s.env.ctx
-	payload := []byte("my data to sign")
+	payload := "my data to sign"
```
ggarri, line 183: We should stick for the usage of bytes

ggarri (approved):

src/nodes/interceptor/eth_sign.go
darioAnongba: done

src/stores/api/handlers/eth1.go
darioAnongba: done

src/stores/api/types/testutils/faker.go
darioAnongba: lol why is this working

darioAnongba: applied

src/stores/connectors/eth1/utils.go
darioAnongba: applied

tests/acceptance/store_test_eth1.go
darioAnongba: applied

PR #250, opened 2021-08-05
TLS AUTH

### LINKED ISSUE #114: Add Client TLS Authenticator
#### Summary

Implement Authenticator allowing to perform Client TLS authentication

#### Acceptance criteria
KM extracts User Information from the subject field of the client certificate
- [ ] Client certificate authentication is enabled by passing the --auth-tls-ca=SOMEFILE runtime option to KM
- [ ] Username from the common name(CN) (e.g. /CN=alice)
- [ ] Groups from using the certificate's organization O) (e.g. /O=group1)

More info
https://docs.google.com/document/d/1cgfbBb7EJ-CAvkR21k0h4GdEC_vS1XdMrEg-PIPvaak/edit#heading=h.x3e8s0m06yw

6l20:
TLS branch cleaned up

## Fixed Issue(s)
fixes #114 

## Changelog


cmd/flags/http.go
```diff
@@ -18,13 +18,54 @@ func init() {
 	viper.SetDefault(httpHostViperKey, httpHostDefault)
 	_ = viper.BindEnv(httpHostViperKey, httpHostEnv)
 
+	viper.SetDefault(httpsPortViperKey, httpsPortDefault)
+	_ = viper.BindEnv(httpsPortViperKey, httpsPortEnv)
+
+	viper.SetDefault(httpsHostViperKey, httpsHostDefault)
```
ggarri, line 24: Duplicated statement

cmd/flags/http.go
```diff
@@ -18,13 +18,54 @@ func init() {
...
+	_ = viper.BindEnv(httpsHostViperKey, httpsHostEnv)
+
+	viper.SetDefault(tlsServerKeyViperKey, tlsServerKeyDefault)
+	_ = viper.BindEnv(tlsServerKeyViperKey, tlsServerKeyEnv)
+
+	viper.SetDefault(tlsServerCertViperKey, tlsServerCertDefault)
```
ggarri, line 30: duplicated statement

pkg/http/server/server.go
```diff
@@ -13,6 +14,17 @@ func New(cfg *Config) *http.Server {
...
 }
 
+func NewTLS(cfg *Config) *http.Server {
+	server := &http.Server{
+		Addr:        fmt.Sprintf("%s:%d", cfg.TLSHost, cfg.TLSPort),
+		IdleTimeout: cfg.IdleConnTimeout,
+		ReadTimeout: cfg.Timeout,
+	}
+	// This will NOT require peer cert during TLS handshake
+	server.TLSConfig = &tls.Config{ClientAuth: tls.VerifyClientCertIfGiven}
```
ggarri, line 24: I would be expecting config.TLSCert and config.TLSKey as could see from this example code https://gist.github.com/denji/12b3a568f092ab951456

src/auth/authenticator/tls/authenticator.go
```diff
@@ -0,0 +1,37 @@
...
+// CN field -> Username
+// Organization -> Groups
+func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	// extract Certificate info from request if any
+	// let go without error when no cert found
+	if req.TLS == nil || len(req.TLS.PeerCertificates) == 0 {
+		return nil, nil
+	}
+	// first array element is the leaf
+	clientCert := req.TLS.PeerCertificates[0]
```
ggarri, line 30: You are missing the checks on the CA passed as argument

tests/e2e/auth_test.go
```diff
@@ -45,6 +49,64 @@ func (t *apiKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		Curve:            "secp256k1",
+		SigningAlgorithm: "ecdsa",
+		PrivateKey:       ecdsaPrivKey,
+		Tags: map[string]string{
+			"myTag0": "tag0",
+			"myTag1": "tag1",
+		},
+	}
+	cert, err := tls.LoadX509KeyPair("certs/same.crt", "certs/common.key")
+	if err != nil {
```
ggarri, line 78: replace by `require.NoError(s.T(), err)`

tests/e2e/auth_test.go
```diff
@@ -45,6 +49,64 @@ func (t *apiKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return http.DefaultTransport.RoundTrip(req)
 }
 
+func TestKeyManagerKeys(t *testing.T) {
```
ggarri, line 52: I was expecting 3 different tests scenarios to validate TLS auth works as expected:
1. Using not certificate should not be rejected (Done)
2. Using valid certificate should be accepted (Done)
3. Using invalid certificate should be rejected (Missing)

pkg/http/server/server.go
```diff
@@ -13,6 +14,17 @@ func New(cfg *Config) *http.Server {
 	}
 }
 
+func NewTLS(cfg *Config) *http.Server {
```
ggarri, line 17: I cannot find any usage of this method. I supposed if the use enables should specify whether or not he want to enable it depending on the values over cmd flags

src/auth/authenticator/tls/authenticator.go
6l20: I was doing a cert comparison here but removed it as we said that we would trust any cert passing the handshake ??

cmd/flags/http.go
6l20: I see no duplication

cmd/flags/http.go
6l20: I see no duplication

pkg/http/server/server.go
6l20: I have missed the TLS server part while picking, I will fix that

ggarri: We shouldn't since https server and auth might use different Authorities to validate certs and we mentioned we will `ClientAuth` customizable in the future

src/auth/authenticator/tls/authenticator.go
ggarri: I would expect something like this:
```
func main() {
  addr := flag.String("addr", ":4000", "HTTPS network address")
  certFile := flag.String("certfile", "cert.pem", "certificate PEM file")
  keyFile := flag.String("keyfile", "key.pem", "key PEM file")
  clientCertFile := flag.String("clientcert", "clientcert.pem", "certificate PEM for client authentication")
  flag.Parse()

  mux := http.NewServeMux()
  mux.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
    if req.URL.Path != "/" {
      http.NotFound(w, req)
      return
    }
    fmt.Fprintf(w, "Proudly served with Go and HTTPS!")
  })

  // Trusted client certificate.
  clientCert, err := os.ReadFile(*clientCertFile)
  if err != nil {
    log.Fatal(err)
  }
  clientCertPool := x509.NewCertPool()
  clientCertPool.AppendCertsFromPEM(clientCert)

  srv := &http.Server{
    Addr:    *addr,
    Handler: mux,
    TLSConfig: &tls.Config{
      MinVersion:               tls.VersionTLS13,
      PreferServerCipherSuites: true,
      ClientCAs:                clientCertPool,
      ClientAuth:               tls.RequireAndVerifyClientCert,
    },
  }

  log.Printf("Starting server on %s", *addr)
  err = srv.ListenAndServeTLS(*certFile, *keyFile)
  log.Fatal(err)
}
```
source https://eli.thegreenplace.net/2021/go-https-servers-with-tls/

tests/e2e/auth_test.go
6l20: Missing bit when I picked from the old branch

pkg/http/server/server.go
6l20: This bit has been missing too

darioAnongba (approved):

Issue #103, opened 2021-05-27
Enforce soft-delete + destroy for all keyVault


ggarri:
The goal of this ticket is to ensure keys are soft-deleted before destroy on every supported key-vault

Issue #237, opened 2021-07-29
Support for RBAC


ggarri:
The goal of this ticket is to support RBAC 

## Acceptance criterial
- [x] Add `allowed_tenants` list to Nodes
- [x] Add `allowed_tenants` list to Stores
- [x] Populate `userInfo` with tenant claims. On every Auth support method (OIDC, ApiKey...) we should extract tenant our to the username field which would have the following proposed format: `{tenant}|{username}`
- [x] Load from manifest a list roles and its permissions
- [x] Implement a authorization component to allow/deny users operations (RBAC version of the policyResolver)

## More info 
https://docs.google.com/document/d/1hRJGpcJr6-hc_tHzb78RXaMjKpghhzC3m0xdIGikye8/edit?usp=sharing

PR #253, opened 2021-08-10
231 db integration

ggarri:
## PR description
- Persist secrets, keys and accounts in DB
- Ensure correct filtering by storeID
- Fix bug on restore secrets and keys
- Fix bug on not selected API-KEY file
- Fix bug on invalid API-KEY authentication
- Refactor nodes folder structure
- Implement basic migration CLI

## Fixed Issue(s)
#231 
#103 

## Changelog


dolanor (dismissed):
First batch of comments on my review (28/156 files reviewed).
I need more, but I have to stop now.

cmd/flags/auth.go
```diff
@@ -221,10 +221,14 @@ func issuerCertificates(vipr *viper.Viper) ([]*x509.Certificate, error) {
 func apiKeyCsvFile(vipr *viper.Viper) (map[string]apikey.UserNameAndGroups, error) {
 	// Open the file
 	csvFileName := vipr.GetString(authAPIKeyFileViperKey)
+	if csvFileName == "" {
+		return nil, nil
+	}
 	csvfile, err := os.Open(csvFileName)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot read api-key filepath %s. %s", csvFileName, err.Error())
```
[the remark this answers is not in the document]
ggarri: Accepeted the suggestion. But I don't get what you mean with wrap the error. This error won't be reused anywhere else as far as I can forseen

cmd/flags/manifest.go
```diff
@@ -37,10 +37,7 @@ func newManifestsConfig(vipr *viper.Viper) (*manifestsmanager.Config, error) {
 	manifestPath := vipr.GetString(manifestPathKey)
 	_, err := os.Stat(manifestPath)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, fmt.Errorf("cannot read manifest path %s", manifestPath)
-		}
-		return nil, err
+		return nil, fmt.Errorf("cannot read manifest path %s. %s", manifestPath, err.Error())
```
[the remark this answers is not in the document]
ggarri: Accepeted the suggestion. But I don't get what you mean with wrap the error. This error won't be reused anywhere else as far as I can forseen

docker-compose.yml
```diff
@@ -33,6 +33,16 @@ x-qkm-common: &qkm-common
   DB_HOST: postgres
 
 services:
+  migration:
```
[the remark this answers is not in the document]
ggarri: As I mentioned above this PR only intends to cover the very basics of the db migration and wait for an spefici PR to cover a more advance version. Are you fine with postpone it till then?

src/app.go
```diff
@@ -56,7 +53,7 @@ func New(cfg *Config, logger log.Logger) (*app.App, error) {
 		return nil, err
 	}
 
-	err = stores.RegisterService(a, logger.WithComponent("stores"), db)
```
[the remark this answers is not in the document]
ggarri: As it was implemented each configs and services are attached to the app in a indenpend way so I applied that principle here. Now DB is being intialized under the initialziation process of the stores service so it decouples each "domain" entirely from each other by not sharing same PG client
```
err = a.RegisterServiceConfig(&stores.Config{Postgres: cfg.Postgres})
```

src/auth/authenticator/api-key/authenticator.go
```diff
@@ -40,6 +40,10 @@ func NewAuthenticator(cfg *Config) (*Authenticator, error) {
 // ? -> Username
 // ? -> Groups
 func (authenticator Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error) {
+	if req.Header.Get("Authorization") == "" {
```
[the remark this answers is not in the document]
ggarri: That is exactly what it happens:
```
	// Authenticate request
	info, err := mid.authenticator.Authenticate(req)
	if err != nil {
		mid.onError(rw, req, err)
		return
	}

	if info != nil {
		// If authentication succeeded then sets the system:authenticated group
		info.Groups = append(types.AuthenticatedUser.Groups, info.Groups...)
	} else {
		// If no authentication then sets info to anonymous user
		info = types.AnonymousUser
	}
```

src/stores/api/formatters/secrets.go
```diff
@@ -2,15 +2,15 @@ package formatters
...
+	"github.com/consensys/quorum-key-manager/src/stores/entities"
 )
 
 func FormatSecretResponse(secret *entities.Secret) *types.SecretResponse {
 	return &types.SecretResponse{
 		ID:          secret.ID,
+		Version:     secret.Version,
 		Value:       secret.Value,
 		Tags:        secret.Tags,
-		Version:     secret.Metadata.Version,
```
dolanor, line 13: What about this metadata field, do we still need it?

cmd/flags/manifest.go
```diff
@@ -37,10 +37,7 @@ func newManifestsConfig(vipr *viper.Viper) (*manifestsmanager.Config, error) {
 	manifestPath := vipr.GetString(manifestPathKey)
 	_, err := os.Stat(manifestPath)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, fmt.Errorf("cannot read manifest path %s", manifestPath)
-		}
-		return nil, err
+		return nil, fmt.Errorf("cannot read manifest path %s: %v", manifestPath, err)
```
dolanor, line 40: As we discussed, we maybe don't need to wrap; but in the other case, you wrap.
We should decide on 1. I guess not wrapping is fine as it's really at app startup that it happens.

src/stores/connectors/eth1/create.go
```diff
@@ -0,0 +1,28 @@
...
+func (c Connector) Create(ctx context.Context, id string, attr *entities.Attributes) (*entities.ETH1Account, error) {
+	logger := c.logger.With("id", id)
+	logger.Debug("creating ethereum account")
+
+	key, err := c.store.Create(ctx, id, eth1Algo, attr)
+	if err != nil {
+		return nil, err
+	}
+	
+	acc, err := c.db.Add(ctx, parseKey(key, attr))
```
dolanor, line 19: What happens if we lose connection to the DB? How do we handle these situations?

src/stores/connectors/eth1/info.go
```diff
@@ -0,0 +1,12 @@
+package eth1
+
+import (
+	"context"
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/consensys/quorum-key-manager/src/stores/entities"
+)
+
+func (c Connector) Info(_ context.Context) (*entities.StoreInfo, error) {
```
dolanor, line 10: Do we have an issue about implementing this?

src/stores/connectors/eth1/restore.go
```diff
@@ -0,0 +1,38 @@
...
+	logger := c.logger.With("address", addr)
+	logger.Debug("restoring ethereum account")
+
+	acc, err := c.db.GetDeleted(ctx, addr)
+	if err != nil {
+		return err
+	}
+
+	err = c.db.RunInTransaction(ctx, func(dbtx database.ETH1Accounts) error {
+		derr := c.db.Restore(ctx, acc)
```
dolanor, line 20: is there a reason for using a different error variable here?

src/stores/connectors/eth1/recover.go
```diff
@@ -0,0 +1,20 @@
+package eth1
+
+import (
+	"context"
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+func (c Connector) ECRevocer(_ context.Context, data, sig []byte) (string, error) {
```
dolanor, line 10:
```suggestion
func (c Connector) ECRecover(_ context.Context, data, sig []byte) (string, error) {
```

src/stores/connectors/eth1/sign.go
```diff
@@ -0,0 +1,294 @@
...
+	quorumtypes "github.com/consensys/quorum/core/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/signer/core"
+	"golang.org/x/crypto/sha3"
+)
+
+var (
```
dolanor, line 22: move them in malleabilityECDSASignature() body since they're only used there?

src/stores/connectors/eth1/sign.go
```diff
@@ -0,0 +1,294 @@
...
+
+	return fmt.Sprintf("\x19\x01%s%s", domainSeparatorHash, typedDataHash), nil
+}
+
+// Azure generates ECDSA signature that does not prevent malleability
+// A malleable signature can be transformed into a new and valid one for a different message or key.
+// https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys-details
+// More info about the issue: http://coders-errand.com/malleability-ecdsa-signatures/
+// More info about the fix: https://en.bitcoin.it/wiki/BIP_0062
+func malleabilityECDSASignature(signature []byte) []byte {
```
dolanor, line 282: If I understand right, this function makes sure the signature is not malleable?
Can we rename to something more sensible? (makeSignatureNonMalleable() ?)

src/stores/connectors/eth1/update.go
```diff
@@ -0,0 +1,41 @@
...
+	logger.Debug("updating ethereum account")
+
+	acc, err := c.db.Get(ctx, addr)
+	if err != nil {
+		return nil, err
+	}
+	acc.Tags = attr.Tags
+
+	err = c.db.RunInTransaction(ctx, func(dbtx database.ETH1Accounts) error {
+		var derr error
```
dolanor, line 22: why derr here?

src/stores/connectors/eth1/verify.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/ethereum/go-ethereum/signer/core"
+)
+
+func (c Connector) Verify(ctx context.Context, addr string, data, sig []byte) error {
+	recoveredAddress, err := c.ECRevocer(ctx, data, sig)
```
dolanor, line 11: ECRecover?

src/stores/connectors/secrets/delete.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/consensys/quorum-key-manager/src/stores/database"
+)
+
+func (c Connector) Delete(ctx context.Context, id, version string) error {
+	logger := c.logger.With("id", id).With("version", version)
+
+	err := c.db.RunInTransaction(ctx, func(dbtx database.Secrets) error {
+		var err error
```
dolanor, line 14: Why a dedicated step?

src/stores/api/formatters/secrets.go
ggarri: Actually it is needed, my mistake was to add "Version" as an attribute in the Secret entity which already have Metadata for it. I am gonna fix that now

src/stores/connectors/eth1/eth1.go
```diff
@@ -1,299 +1,46 @@
...
+func NewEth1Connector(store stores.KeyStore, db database.ETH1Accounts, logger log.Logger) *Connector {
+	return &Connector{
+		store:  store,
+		logger: logger,
+		db:     db,
+	}
+}
+
+func parseKey(key *entities.Key, attr *entities.Attributes) *entities.ETH1Account {
+	pubKey, _ := crypto.UnmarshalPubkey(key.PublicKey)
```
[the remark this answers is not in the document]
ggarri: This is being a long discussion with @darioAnongba . The conclusion is we ignore errors in case we control the data flow and we can guarantee no error can occur. In this case the PublicKey is something we already validated on the infra layer so that at the time it reaches this part of the code we can ignore the error checks

src/stores/connectors/eth1/info.go
ggarri: We just didnt implemented yet

src/stores/connectors/eth1/create.go
ggarri: Create an issue to discuss best way to deal with this situation

src/stores/connectors/eth1/info.go
ggarri: Creata ticket to implement info() method

src/stores/connectors/eth1/sign.go
ggarri: I didn't modify this in this PR so lets not modify it here

src/stores/connectors/eth1/update.go
ggarri: shadowing issue in linter

src/stores/connectors/eth1/sign.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	logger.Debug("signed hashed payload successfully")
+	return signature, nil
+}
+
+func (c Connector) SignTypedData(ctx context.Context, addr string, typedData *core.TypedData) ([]byte, error) {
+	logger := c.logger.With("address", addr)
+
+	encodedData, err := getEIP712EncodedData(typedData)
```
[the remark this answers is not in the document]
ggarri: I didn't modify this in this PR so lets not modify it here

src/stores/connectors/secrets/delete.go
ggarri: fixed

src/stores/connectors/eth1/create.go
ggarri: https://github.com/ConsenSys/quorum-key-manager/issues/256

src/stores/connectors/eth1/sign.go
ggarri: This code does not belong to my PR since it was only moved from `/src/stores/store/eth1` to here

src/stores/manager/service.go
```diff
@@ -125,64 +121,58 @@ func (m *BaseManager) loadAll() {
...
 	}
 
 	errMessage := "secret store was not found"
-	m.logger.Error(errMessage, "store_name", name)
+	m.logger.Error(errMessage, "store_name", storeName)
 	return nil, errors.NotFoundError(errMessage)
 }
 
-func (m *BaseManager) GetKeyStore(ctx context.Context, name string, userInfo *authtypes.UserInfo) (connectors.KeysConnector, error) {
+func (m *BaseManager) GetKeyStore(_ context.Context, storeName string, _ *authtypes.UserInfo) (stores.KeyStore, error) {
```
dolanor, line 141: don't we need the context for the connector? What about the user info?

src/stores/manager/service.go
```diff
@@ -125,64 +121,58 @@ func (m *BaseManager) loadAll() {
...
+func (m *BaseManager) getEth1Store(ctx context.Context, storeName string, userInfo *authtypes.UserInfo) (stores.Eth1Store, error) {
+	if storeBundle, ok := m.eth1Accounts[storeName]; ok {
+		if store, ok := storeBundle.store.(stores.KeyStore); ok {
 			policies := m.policyManager.UserPolicies(ctx, userInfo)
-			resolvr, err := policy.NewResolver(policies)
-			if err != nil {
-				return nil, err
-			}
-			return eth1.NewEth1Connector(store, resolvr, storeBundle.logger), nil
+			_, _ = policy.NewResolver(policies)
```
dolanor, line 165: why we don't care about what NewResolver does?

src/stores/store/keys/akv/akv.go
```diff
@@ -91,15 +93,46 @@ func (s *Store) Import(ctx context.Context, id string, privKey []byte, alg *enti
...
+		kID, _ := parseKeyID(kItem.Kid)
+		kIDs = append(kIDs, kID)
+	}
+
+	return kIDs, nil
+}
+
 func (s *Store) Update(ctx context.Context, id string, attr *entities.Attributes) (*entities.Key, error) {
-	res, err := s.client.UpdateKey(ctx, id, "", nil, nil, attr.Tags)
+	expireAt := date.NewUnixTimeFromNanoseconds(time.Now().Add(attr.TTL).UnixNano())
```
dolanor, line 132: can't we just use time.Time() for that?

dolanor: Forget about it, it is necessary for AKV, for whatever reason…

src/stores/manager/service.go
ggarri: since there is not async calls made within the execution of this method and logging does not depend on ctx, there is not need for context. We added context to all our methods just for consistency and to prevent method signature changes in case context is required in the future

src/stores/manager/service.go
ggarri: PolicyResolver will be implemented as part of this ticket #237

darioAnongba (approved):

PR #254, opened 2021-08-11
Alias API

### LINKED ISSUE #147: Implement API allowing to manage aliases
#### Summary

Implement API for alias service allow to manage aliases

#### Acceptance criteria

- [ ] Add `POST /registries/{name}/aliases/{alias}`
- [ ] Add `GET /registries/{name}/aliases/{alias}`
- [ ] Add `DELETE /registries/{name}/aliases/{alias}`
- [ ] Add `PATCH /registries/{name}/aliases/{alias}`
- [ ] Add `GET /registries/{name}`
- [ ] Add `DELETE /registries/{name}`
- [ ] Add an HTTP Client to access above routes
- [ ] Add Acceptance tests
- [ ] Add swagger documentation

dolanor:
Fixes #147 partially (no acceptance tests yet, will come after the DB migration PR)


deps/config/default.yml
```diff
@@ -0,0 +1,95 @@
+kind: AKVKeys
```
ggarri, line 1: Manifest template files are updated under `./deps/config/manifest/*.template`. Are you intending to change?

src/aliases/api/api_test.go
```diff
@@ -0,0 +1,397 @@
+package aliasapi_test
```
ggarri, line 1: Following what we did in `/stores/api` unit test are being written over handlers file so we test one by one each endpoint and each edge case. Having them all together in a single file it makes things harder to maintain and follow

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,166 @@
...
+
+	alRoute := regRoute.PathPrefix("/aliases").Subrouter()
+	alRoute.HandleFunc("", h.createAlias).Methods(http.MethodPost)
+	alRoute.HandleFunc("", h.listAliases).Methods(http.MethodGet)
+	alRoute.HandleFunc("/{alias_key}", h.getAlias).Methods(http.MethodGet)
+	alRoute.HandleFunc("/{alias_key}", h.updateAlias).Methods(http.MethodPut)
+	alRoute.HandleFunc("/{alias_key}", h.deleteAlias).Methods(http.MethodDelete)
+}
+
+func (h *AliasHandler) deleteRegistry(w http.ResponseWriter, r *http.Request) {
```
ggarri, line 38: Please, remember to include endpoints annotations docs as you can see in `/stores/api/handlers/eth1.go` 
```
// @Summary Creates a secret
// @Description Create new secret on selected Store
// @Tags Secrets
// @Accept json
// @Produce json
// @Param id path string true "Secret Identifier"
// @Param storeName path string true "Store Identifier"
// @Param request body types.SetSecretRequest true "Create Secret request"
// @Success 200 {object} types.SecretResponse "Secret data"
// @Failure 400 {object} ErrorResponse "Invalid request format"
// @Failure 404 {object} ErrorResponse "Store not found"
// @Failure 500 {object} ErrorResponse "Internal server error"
// @Router /stores/{storeName}/secrets/{id} [post]
```

src/aliases/api/handlers/utils.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+)
+
+func jsonWrite(w http.ResponseWriter, data interface{}) error {
```
ggarri, line 12: To return a json object in the http response we are using the following

```
import "encoding/json"
...
_ = json.NewEncoder(rw).Encode(formatters.FormatSecretResponse(secret))
```

src/aliases/api/handlers/utils.go
```diff
@@ -0,0 +1,68 @@
...
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8;")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	return json.NewEncoder(w).Encode(data)
+}
+
+func WriteHTTPErrorResponse(rw http.ResponseWriter, err error) {
```
ggarri, line 18: We shouldn't duplicate this method (we already have the same in /stores/api/...`). Instead we should have all comoon http common stuff under `/src/infra/http` for reusability. `infra` is intended to encapsulate every implementation required to communicate (inbound and outbound) with external services and this includes http

src/aliases/api/handlers/utils.go
```diff
@@ -0,0 +1,68 @@
...
+		writeErr = writeErrorResponse(rw, http.StatusInternalServerError, fmt.Errorf(internalErrMsg))
+	}
+	if writeErr != nil {
+		// TODO the: use logger
+		log.Printf("error writing the original error: %v: %v", writeErr, err)
+		http.Error(rw, writeErr.Error(), http.StatusInternalServerError)
+	}
+}
+
+func writeErrorResponse(w http.ResponseWriter, status int, err error) error {
```
ggarri, line 43: Same about moving common http code to `/src/infra/http` and reuse it from there

src/aliases/api/types/alias.go
```diff
@@ -0,0 +1,61 @@
+package types
+
+import aliasent "github.com/consensys/quorum-key-manager/src/aliases/entities"
+
+type Alias struct {
```
ggarri, line 5: Following `/nodes` convention we should use naming such as `AliasRequest` and `AliasResponse` in addition we would have `formatters` to parse Request/Response from/to Entities

src/aliases/api/types/alias.go
```diff
@@ -0,0 +1,61 @@
...
+	Key   AliasKey   `json:"key"`
+	Value AliasValue `json:"value"`
+
+	registryName RegistryName
+}
+
+func FromEntityAlias(ent aliasent.Alias) Alias {
```
ggarri, line 12: In order to be more consistant we should have `formatters` folder to encapsulate Request/Response from/to Entities logic

src/aliases/api/types/alias.go
ggarri, line 5: as far as I see `Alias` is defined for reusability, so what you expose here are the Request and Response structs. Therefore I would suggest to make this stuct not exposed. In addition when you define you Request and Response would should define them with a pointer so you don't end up with nested attributes instead of flat object

```
type CreateAliasRequest struct {
	*Alias
}
```

src/aliases/api/handlers/handlers.go
dolanor: Completely forgot about that.

src/aliases/api/handlers/utils.go
dolanor: It's mostly to have a common way to add the HTTP headers homogenically.
In the end, it just does what you say as well.

src/aliases/api/handlers/utils.go
dolanor: I understand that, but is it the point of this PR to move all this code into a common area and update code from stores to use the new common package?

src/aliases/api/types/alias.go
dolanor: Why can't we keep it in the types package? It's type transformation from/to domain entities.
Domain entities shouldn't know about other format, but the other way around is good. Therefore, I don't see the point of splitting even more packages when the subject of those formatters would be those types. And those formatters without those types would have no purpose at all.

src/aliases/api/types/alias.go
dolanor: Can we have a call about this, I'm not sure I'm getting everything from this :)

src/aliases/manager/manager.go
```diff
@@ -0,0 +1,50 @@
+package aliasmanager
```
ggarri, line 1: I cannot understand the propose of having a AliasManager. Managers are long living instances of our application which are intended for holding a pice of the application state. For instance in the case of nodes and stores, the manager is notified about manifest content in order to instanciate Stores and Nodes services which are the ones connected to the DB and KeyVaults but the manager never interacts with any of those directly as it is your case here.

In your case, since you don't have any state to be held I would rather propose to initialize an AliasConnector (just to follow same convention than Stores) instead of the manager and that service is the one being passed to your handlers to interact with the db

src/aliases/manager/service.go
```diff
@@ -1,50 +1,19 @@
...
-type BaseManager struct {
-	db aliasstore.Database
-}
-
-func New(db aliasstore.Database) *BaseManager {
-	return &BaseManager{
-		db: db,
+// RegisterService creates and register the alias service in the app.
+func RegisterService(a *app.App, pgClient postgres.Client) error {
+	db := aliaspg.NewDatabase(pgClient)
```
ggarri, line 11: As we are discussing on the other PR(https://github.com/ConsenSys/quorum-key-manager/pull/253/files#diff-f6c71e04a55bbc1dbf8f31101336e0ac7589dabdd7a654c60a00964281bd0443R22) I would rather create a new client for each domain so we keep the intended independency between services

src/aliases/manager/service.go
```diff
@@ -1,50 +1,19 @@
...
 
-type BaseManager struct {
-	db aliasstore.Database
-}
-
-func New(db aliasstore.Database) *BaseManager {
-	return &BaseManager{
-		db: db,
+// RegisterService creates and register the alias service in the app.
+func RegisterService(a *app.App, pgClient postgres.Client) error {
```
ggarri, line 10: Following conventions from `nodes` and `stores`, this code belongs to `/src/aliases/service.go`

src/aliases/store/postgres/alias.go
```diff
@@ -73,5 +73,5 @@ func (s *AliasStore) ListAliases(ctx context.Context, registry aliasent.Registry
 }
 
 func (s *AliasStore) DeleteRegistry(ctx context.Context, registry aliasent.RegistryName) error {
-	return goerrors.New("not implemented")
+	return errors.NotImplementedError("DeleteRegistry not implemented")
```
ggarri, line 76: As part of the common practice over our code base, we should be logging error happens, EXCEPT when you are in `pkg/` which acts as libraries, and in that case we log on the first component under `/src`

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return &h
+}
+
+func (h *AliasHandler) Register(r *mux.Router) {
+	regRoute := r.PathPrefix("/registries/{registry_name}").Subrouter()
+	regRoute.HandleFunc("", h.deleteRegistry).Methods(http.MethodDelete)
+
+	alRoute := regRoute.PathPrefix("/aliases").Subrouter()
+	alRoute.HandleFunc("", h.createAlias).Methods(http.MethodPost)
```
ggarri, line 31: To be consistent with `/stores/api` you should include `alias_key` in the url

deps/config/default.yml
dolanor: repinging @nmvalera for this specific topic.

dolanor: Confirmed, I'll remove this.

src/aliases/manager/service.go
dolanor: done

src/aliases/manager/service.go
dolanor: In stores, that code is in `/src/stores/app/service.go`. For `nodes` it is directly in the main directory.
I'll do it like that.

src/aliases/manager/manager.go
dolanor: I need to have a Runnable to register it, even though it doesn't deal with any state.
Saw that suggestion with @nmvalera and we will keep as is.

src/aliases/api/types/alias.go
dolanor: I don't feel the name `FormatXxx` does convey as much information as `FromXxx` and `ToXxx`.
In `stores/api/formatters`, some `FormatXxx` means it will transform to `Xxx`, and in other cases, `FormatXxx` means we transform `Xxx` into something else.
With From and To, we transform `From` a type from another package to the current one, and `To` would transform a current package type to another one. I think the name conveys directly more meaning without needing to read the whole signature and return types.

src/aliases/api/handlers/utils.go
dolanor: done already in another PR.

src/aliases/api/handlers/utils.go
dolanor: done already in another PR.

src/aliases/api/api_test.go
dolanor: done

src/aliases/manager/service.go
dolanor: done

src/aliases/api/types/alias.go
dolanor: I'll keep this suggestion in our future discussion, I'll rename like in `stores/api/formatters` for now.

dolanor: done

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+	h := AliasHandler{
+		alias: backend,
+	}
+
+	return &h
+}
+
+func (h *AliasHandler) Register(r *mux.Router) {
+	regRoute := r.PathPrefix("/registries/{registry_name}").Subrouter()
+	regRoute.HandleFunc("", h.deleteRegistry).Methods(http.MethodDelete)
```
ggarri, line 31: Isn't it possible to list all available registries?

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+// @Summary Delete a registry
+// @Description Delete a registry and all its keys
+// @Tags registry
+// @Param registry_name path string true "registry identifier"
+// @Param registry_key path string true "registry identifier"
+// @Param request body types.DeleteRegistryRequest true "Delete Registry Request"
+// @Success 204 {object} types.DeleteRegistryResponse "Registry data"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
+// @Failure 500 {object} ErrorResponse "Internal server error"
+// @Router /registryes/{registry_name}/registryes/{registry_key} [delete]
```
ggarri, line 50: typo, plural of registry is `registries`

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+	alRoute.HandleFunc("/{alias_key}", h.getAlias).Methods(http.MethodGet)
+	alRoute.HandleFunc("/{alias_key}", h.updateAlias).Methods(http.MethodPut)
+	alRoute.HandleFunc("/{alias_key}", h.deleteAlias).Methods(http.MethodDelete)
+}
+
+// @Summary Delete a registry
+// @Description Delete a registry and all its keys
+// @Tags registry
+// @Param registry_name path string true "registry identifier"
+// @Param registry_key path string true "registry identifier"
```
ggarri, line 45: What is the difference between `registry_key` and `registry_name` ? I cannot see any usage of `registry_key` in the below implementation

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+		infrahttp.WriteHTTPErrorResponse(w, errors.InvalidFormatError(err.Error()))
+		return
+	}
+
+	err = h.alias.DeleteRegistry(r.Context(), aliasent.RegistryName(regName))
+	if err != nil {
+		infrahttp.WriteHTTPErrorResponse(w, err)
+		return
+	}
+	w.Header().Set("X-Content-Type-Options", "nosniff")
```
ggarri, line 67: What is this header for?

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+// @Tags alias
+// @Accept json
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.CreateAliasRequest true "Create Alias Request"
+// @Success 200 {object} types.CreateAliasResponse "Alias data"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
+// @Failure 500 {object} ErrorResponse "Internal server error"
+// @Router /aliases/{registry_name}/aliases/{alias_key} [post]
```
ggarri, line 82: Invalid Router, it should be:
```
@Router /registries/{registry_name}/aliases/{alias_key} [post]
```

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+	alias, err := h.alias.CreateAlias(r.Context(), eAlias.RegistryName, eAlias)
+	if err != nil {
+		infrahttp.WriteHTTPErrorResponse(w, err)
+		return
+	}
+
+	resp := types.AliasResponse{
+		Value: types.AliasValue(alias.Value),
+	}
+	err = jsonWrite(w, resp)
```
ggarri, line 112: As we discussed `jsonWrite` method should be moved into `/src/infra/http`

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+// @Summary Get an alias
+// @Description Get an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Success 200 {object} types.GetAliasResponse "Alias data"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
+// @Failure 500 {object} ErrorResponse "Internal server error"
+// @Router /aliases/{registry_name}/aliases/{alias_key} [get]
```
ggarri, line 128: Invalid router

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+// @Summary Delete an alias
+// @Description Delete an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.DeleteAliasRequest true "Delete Alias Request"
+// @Success 204 "Deleted successfully"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
+// @Failure 500 {object} ErrorResponse "Internal server error"
+// @Router /aliases/{registry_name}/aliases/{alias_key} [delete]
```
ggarri, line 222: Invalid router

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+
+// deleteAlias deletes an alias value.
+// @Summary Delete an alias
+// @Description Delete an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.DeleteAliasRequest true "Delete Alias Request"
+// @Success 204 "Deleted successfully"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
```
ggarri, line 220: I believe a 404 is also a possible error

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+// @Summary Update an alias
+// @Description Update an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Accept json
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.UpdateAliasRequest true "Update Alias Request"
+// @Success 200 {object} types.UpdateAliasResponse "Alias data"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
```
ggarri, line 166: I believe a 404 is also a possible error

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// @Summary Get all the aliases in a registry
+// @Description Get all the aliases in a registry
+// @Tags alias
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Success 200 {array} types.GetAliasResponse "a list of Aliases"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
```
ggarri, line 252: is it really possible to send an invalid request format?

src/aliases/api/handlers/utils.go
ggarri: As we agree, we would need to move this into `/src/infra/http`

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+}
+
+// deleteAlias deletes an alias value.
+// @Summary Delete an alias
+// @Description Delete an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.DeleteAliasRequest true "Delete Alias Request"
```
ggarri, line 218: As far as I can see there is not any request body expected

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+
+// updateAlias updates an alias value.
+// @Summary Update an alias
+// @Description Update an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Accept json
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.UpdateAliasRequest true "Update Alias Request"
```
ggarri, line 164: I cannot see any type in your code corresponding to `UpdateAliasRequest`

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+}
+
+// @Summary Get an alias
+// @Description Get an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Success 200 {object} types.GetAliasResponse "Alias data"
```
ggarri, line 125: I cannot see any `GetAliasResponse` in your PR. Did I miss something?

src/aliases/api/handlers/handlers.go
dolanor: Maybe in another PR
cc @nmvalera

src/aliases/api/handlers/handlers.go
dolanor: I copied it from https://github.com/ConsenSys/quorum-key-manager/commit/7a6e202#diff-5ef92eb3fa1b476f02e830de80aea28c596776a886f710278317a67579f66765

TIL about deprecating the `X-` prefixing, thanks!
But in that case, I thought it was mostly to make IE not try to detect charset from the initial bytes, because it can trip for json and thinks the charset is the default ISO-8859-1, although, for application/json, it is specifically utf-8 by default.

src/aliases/api/handlers/handlers.go
dolanor: done

src/aliases/api/handlers/handlers.go
dolanor: it shouldn't be there… Really bad copy paste and reread job on my part.
I removed it.

src/aliases/api/handlers/handlers.go
dolanor: @nmvalera said we should also get rid of it.

src/aliases/api/handlers/handlers.go
dolanor: Yes, in the case the registry_name in the path doesn't respect the format
```go
const pathVarsFormat = "^[a-zA-Z0-9-_+]+$"
```

src/aliases/api/handlers/handlers.go
dolanor: You're right. Removed.

src/aliases/api/handlers/handlers.go
dolanor: Forgot to update this once streamlining the API types

src/aliases/api/handlers/handlers.go
dolanor: Forgot to update this once streamlining the API types

src/aliases/api/handlers/handlers.go
dolanor: Yes!

src/aliases/api/handlers/handlers.go
dolanor: Yes!

src/aliases/api/handlers/handlers.go
dolanor: fixed

src/aliases/api/handlers/handlers.go
dolanor: fixed

src/aliases/api/handlers/handlers.go
dolanor: fixed

src/aliases/api/handlers/handlers.go
ggarri: GTM

src/aliases/api/handlers/handlers.go
ggarri: but in case path does not match I would expect a 404 `Page Not Found` instead of 400. isn't that right?

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,297 @@
...
+	alRoute.HandleFunc("/{alias_key}", h.getAlias).Methods(http.MethodGet)
+	alRoute.HandleFunc("/{alias_key}", h.updateAlias).Methods(http.MethodPut)
+	alRoute.HandleFunc("/{alias_key}", h.deleteAlias).Methods(http.MethodDelete)
+}
+
+// @Summary Delete a registry
+// @Description Delete a registry and all its keys
+// @Tags registry
+// @Param registry_name path string true "registry identifier"
+// @Success 204 {object} types.DeleteRegistryResponse "Registry data"
```
ggarri, line 45: Struct `DeleteRegistryResponse` does not exist. Please, make sure you can run `$> make gen-swagger` using this annotation, because I believe it will not work

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,297 @@
...
+
+// deleteAlias deletes an alias value.
+// @Summary Delete an alias
+// @Description Delete an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Success 204 "Deleted successfully"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
+// @Failure 404 {object} ErrorResponse "Alias not found"
```
ggarri, line 218: In case of registry not found, isn't it a 404 too?

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,297 @@
...
+
+// @Summary Get an alias
+// @Description Get an alias of a key from a dedicated alias registry
+// @Tags alias
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Success 200 {object} types.AliasResponse "Alias data"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
+// @Failure 404 {object} ErrorResponse "Alias not found"
```
ggarri, line 126: In case of registry not found, isn't it a 404 too?

src/aliases/api/handlers/handlers.go
```diff
@@ -0,0 +1,297 @@
...
+// @Summary Creates an alias
+// @Description Create an alias of a key in a dedicated alias registry
+// @Tags alias
+// @Accept json
+// @Produce json
+// @Param registry_name path string true "registry identifier"
+// @Param alias_key path string true "alias identifier"
+// @Param request body types.AliasRequest true "Create Alias Request"
+// @Success 200 {object} types.AliasResponse "Alias data"
+// @Failure 400 {object} ErrorResponse "Invalid request format"
```
ggarri, line 79: You are missing 404

src/aliases/api/handlers/handlers.go
dolanor: We actually don't have a registry, it's part of the same table, and it's the same composed key. So either/or, it will fail.
And since adding a key to an unexisting registry will create the "registry", then it will succeed.
It can either succeed if both ids are good, or it fails if any of them fails, we can't know which was wrong. But a good alias key in the wrong registry doesn't make it good anyway. So just a 404 saying about Alias should be good.

The real response says the resource is unavailable.
```
{"message":"ST100: resource not found","code":"ST100"}
```

src/aliases/api/handlers/handlers.go
dolanor: We actually don't have a registry, it's part of the same table, and it's the same composed key. So either/or, it will fail.
And since adding a key to an unexisting registry will create the "registry", then it will succeed.
It can either succeed if both ids are good, or it fails if any of them fails, we can't know which was wrong. But a good alias key in the wrong registry doesn't make it good anyway. So just a 404 saying about Alias should be good.

The real response says the resource is unavailable.
```
{"message":"ST100: resource not found","code":"ST100"}
```

src/aliases/api/handlers/handlers.go
dolanor: It stops at:
```
2021/09/07 09:28:06 Generate swagger docs....
2021/09/07 09:28:06 Generate general API Info, search dir:./src
2021/09/07 09:28:06 Generating types.CreateKeyRequest
2021/09/07 09:28:06 Generating types.KeyResponse
2021/09/07 09:28:06 Generating entities.Annotation
2021/09/07 09:28:06 Generating handlers.ErrorResponse
2021/09/07 09:28:06 Generating types.ImportKeyRequest
2021/09/07 09:28:06 Generating types.SignBase64PayloadRequest
2021/09/07 09:28:06 Generating types.CreateEth1AccountRequest
2021/09/07 09:28:06 Generating types.Eth1AccountResponse
2021/09/07 09:28:06 Generating types.ImportEth1AccountRequest
2021/09/07 09:28:06 Generating types.UpdateEth1AccountRequest
2021/09/07 09:28:06 Generating types.SignMessageRequest
2021/09/07 09:28:06 Generating types.SignTypedDataRequest
2021/09/07 09:28:06 Generating types.DomainSeparator
2021/09/07 09:28:06 Generating types.Type
2021/09/07 09:28:06 Generating types.SignETHTransactionRequest
2021/09/07 09:28:06 Generating types.SignEEATransactionRequest
2021/09/07 09:28:06 Generating types.SignQuorumPrivateTransactionRequest
2021/09/07 09:28:06 Generating types.ECRecoverRequest
2021/09/07 09:28:06 ParseComment error in file src/stores/api/handlers/eth1.go :cannot find type definition: types.VerifyEth1SignatureRequest
make: *** [Makefile:129: gen-swagger] Error 
```

src/aliases/api/handlers/handlers.go
dolanor: done

src/aliases/manager/manager.go
dolanor: Finally removed it after a discussion with @ggarri.
The service doesn't need to be registered this way as the routes are registered via the mux.Router.

But now, if I need to check Liveness/Readiness checks, I guess I need to re-create a Runnable?
Let's do that in another small PR, shall we?

src/aliases/api/handlers/utils.go
dolanor: done

src/aliases/api/handlers/handlers.go
dolanor: There is no 404 as the registry will be created if it doesn't exist yet (it is just an ID that is part of the key).

src/aliases/api/handlers/handlers.go
dolanor: From https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
> 400 Bad Request
> 
> The HyperText Transfer Protocol (HTTP) 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

I would consider this a `deceptive request routing`.
Given their browser, they maybe want an alias with a space ` `, and the browser could display the url as `/registries/my-registry/aliases/my%20alias`. That could confuse them.

src/aliases/api/handlers/handlers.go
ggarri: this was fixed on my last PR. Please rebase main I would expect following errors to be linked to your Aliases API docs

src/aliases/api/handlers/handlers.go
ggarri: I see, resolving

src/aliases/api/handlers/handlers.go
ggarri: I also get confused by the fact a ID are defined in URl and then you can get a 400 but it is fine I would say

ggarri (approved):

src/aliases/api/handlers/handlers.go
dolanor: done

Issue #223, opened 2021-07-26
Use jwt token standard claim fields as default


darioAnongba:
In the JWT standard,  we have:
- `sub` is used to define of the username and organization. ie `tenant0|username1`
- `scope` is used to define the groups and claims. ie `openid profile email read:current_user`

Access https://jwt.io/ and introduce the following token generated via Auth0
```
[...]

PR #257, opened 2021-08-16
237 rbac

ggarri:
## PR description
- Manifest loader for application Roles
- Defined `allowed_tenants` for Nodes and Stores
- Added support for `tenants` in Authorization middleware
- Restrict access to Nodes and Stores based on user's tenant claims
- Use default values for JWT claims (`sub` for username and tenant, `scope` for roles and permissions)

## Fixed Issue(s)
#237 
#223 

## Changelog


src/auth/manager/operation.go
```diff
@@ -0,0 +1,20 @@
...
+type OpAction string
+type OpResource string
+
+var ActionRead OpAction = "read"
+var ActionWrite OpAction = "write"
+var ActionSign OpAction = "sign"
+var ActionDelete OpAction = "delete"
+var ActionDestroy OpAction = "destroy"
+
+var ResourceKey OpResource = "key"
```
darioAnongba, line 12: I think the resources should be plural. Also isn't `stores` missing?

src/auth/manager/resolver.go
```diff
@@ -0,0 +1,59 @@
...
+	return res.isAllowed
+}
+
+func (res *Result) Error() error {
+	return res.err
+}
+
+// Resolver is responsible to control whether an operation is authorized or not
+// depending on the set of policies attached to the resolver
+type Resolver struct {
```
darioAnongba, line 25: I don't see the need for a `Resolver` at all. It seems that we can simply do a `permissions.Contains("read:keys")` on the connector and it should be enough right? What I mean is that we can use pure functions and do not need a stateful struct.

src/auth/types/user.go
```diff
@@ -0,0 +1,51 @@
...
+	AuthMode string
+
+	// Tenant belonged by the user
+	Tenant string
+
+	// Username identifies the user
+	Username string
+
+	// Roles indicates the user's membership
+	Roles []string
```
darioAnongba, line 20: Do we need Roles here? Can't we translate the role into a list of permissions when creating the UserInfo and delete this field from the struct?

src/auth/types/user.go
```diff
@@ -0,0 +1,51 @@
...
+	Permissions []Permission
+}
+
+func (ui *UserInfo) CheckAccess(mnf *manifest.Manifest) error {
+	if len(mnf.AllowedTenants) == 0 {
+		return nil
+	}
+
+	if ui.Tenant == "" {
+		return errors.NotFoundError("missing credentials")
```
darioAnongba, line 32: I'm not sure it's NotFound here, I think here it's Unauthorized because credentials are missing (401).

src/auth/manager/operation.go
ggarri: You r right, stores are missing. I will also use plurals and correlate that change with the permissions so the match

This vars were still not used since that will be added onces we implement the stores/nodes authorization.

src/auth/types/user.go
ggarri: Roles are converted into Permissions on the authorization middlewares as you mention BUT I think we will need Roles in the UserInfo because that information will be taken by the Audit Component to write records of which user with which roles accessed the app etc.

src/auth/manager/resolver.go
ggarri: You are right at the current version having a Resolver appers to be not very necesiary but IMO this dedicated component help us to decouple out authorization logic out from the Connectors and improve scalability in case we decide on making more advance auth rules such as ownership base rules where the Resolver would need to read from DB data.

darioAnongba (approved):

PR #261, opened 2021-08-25
Authorization checks for nodes

### LINKED ISSUE #122: Add authorization checks on `nodes` service
#### Summary 

Implement authorization check on `nodes` service and `jsonrpc` interceptor

darioAnongba:
## Fixed Issue(s)
fixes #122

### DIFF
--- a/src/nodes/manager/service.go
+++ b/src/nodes/manager/service.go
@@ -6,6 +6,9 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/consensys/quorum-key-manager/src/auth"
+	"github.com/consensys/quorum-key-manager/src/auth/authorizator"
+
 	authtypes "github.com/consensys/quorum-key-manager/src/auth/types"
 	"github.com/consensys/quorum-key-manager/src/infra/log"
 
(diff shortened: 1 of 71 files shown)


src/nodes/manager/service.go
```diff
@@ -115,19 +120,19 @@ func (m *BaseManager) Node(_ context.Context, name string, userInfo *authtypes.U
...
+			return nil, err
 		}
-		return nodeBundle.node, nodeBundle.err
-	}
 
-	// This piece of code is here to make sure it is possible to retrieve a default node
-	for _, nodeBundle := range m.nodes {
-		if err := userInfo.CheckAccess(nodeBundle.manifest); err != nil {
-			continue
+		err = resolver.CheckPermission(&authtypes.Operation{Action: authtypes.ActionProxy, Resource: authtypes.ResourceNode})
```
ggarri, line 131: [IMPORTANT] You cannot check `proxy:node` permission here as you don't know whether or not the node will be use to proxy a call when you call this method Node(...). Instead the should be check at the Interceptor component where the proxy action really happens

ggarri (approved):

src/nodes/manager/service.go
ggarri: Since current implementation is pending of a refactor to include some intermedidate components to allow us to create fine grane permissions checks we decided over keeping this appoach as good for now.

PR #276, opened 2021-08-30
PostgreSQL Migrations

darioAnongba:
Implement PostgreSQL migrations.

Missing:
- Migrations running using SSL mode
- Migrations "down" command line


cmd/flags/migration.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func init() {
+	viper.SetDefault(MigrationSourceURLViperKey, migrationSourceURLDefault)
+	_ = viper.BindEnv(MigrationSourceURLViperKey, migrationSourceURLENV)
```
ggarri, line 11: do we need flags to specify where migrations are located? I cannot forseen users using differenet migration folder than the default one. If you cannot find specific use-cases for it I would suggest to remove this to prevent us to maintain it

darioAnongba: removed.

cmd/migrate.go
```diff
@@ -44,28 +44,19 @@ func migrateCmd(cmd *cobra.Command, _ []string) error {
 	}
 	defer syncZapLogger(logger)
 
-	pgOpt, err := pgCfg.ToPGOptions()
+	dbURL := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=disable", pgCfg.User, pgCfg.Password, pgCfg.Host, pgCfg.Port, pgCfg.Database)
```
ggarri, line 47: We have PG Client Dialer implementation has a very advance behaviour which allow us to support many different SSL modes and it is already integrated with out cmd flags. Said that, I would rather investtime in trying to reuse that PG Client into this migration instance, most likely you will need to implement the Driver interface, but IMO we cannot get along without it in this PR

deps/config/migrations/000001_create_tables.down.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
```
ggarri, line 1: Migrations are not config, therefore I would rather place them un level un, just in `deps/migration` or even at the root `/migrations`

deps/config/migrations/000001_create_tables.up.sql
```diff
@@ -0,0 +1,51 @@
...
+    updated_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL,
+    deleted_at TIMESTAMPTZ,
+    PRIMARY KEY (id, version, store_id)
+);
+
+CREATE TABLE IF NOT EXISTS keys (
+    id VARCHAR (50) NOT NULL,
+    store_id VARCHAR (50) NOT NULL,
+    public_key BYTEA NOT NULL,
+    signing_algorithm VARCHAR (50) NOT NULL,
```
ggarri, line 19: There is not additional value in using VARCHARs onto TEXT, therefore I would rather replace all VARCHAR to TEXT to prevent us for doing wrong assumption about field sizes

deps/config/migrations/000001_create_tables.up.sql
```diff
@@ -0,0 +1,51 @@
...
+CREATE TABLE IF NOT EXISTS secrets (
+    id VARCHAR (50) NOT NULL,
+    version VARCHAR (50) NOT NULL,
+    store_id VARCHAR (50) NOT NULL,
+    disabled BOOLEAN default false,
+    tags JSONB,
+    created_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL,
+    updated_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL,
+    deleted_at TIMESTAMPTZ,
+    PRIMARY KEY (id, version, store_id)
```
ggarri, line 12: Compose PRIMARY KEY are very expensive in DBs, therefore I would rather create a autoincrement PK and then an UNIQUE (id, version, store_id)

deps/config/migrations/000001_create_tables.up.sql
```diff
@@ -0,0 +1,51 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS keys (
+    id VARCHAR (50) NOT NULL,
```
ggarri, line 16: String PRIMARY KEYs are very expensive in DBs, therefore I would rather create a autoincrement `pk` and then an UNIQUE (id)

deps/config/migrations/000001_create_tables.up.sql
```diff
@@ -0,0 +1,51 @@
...
+    store_id VARCHAR (50) NOT NULL,
+    key_id VARCHAR (50) NOT NULL,
+    public_key BYTEA NOT NULL,
+    compressed_public_key BYTEA NOT NULL,
+    tags JSONB,
+    disabled BOOLEAN default false,
+    created_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL,
+    updated_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL,
+    deleted_at TIMESTAMPTZ,
+    PRIMARY KEY (address, store_id)
```
ggarri, line 41: Same comment about compose primary keys

deps/config/migrations/000001_create_tables.up.sql
```diff
@@ -0,0 +1,51 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS aliases (
+    key VARCHAR (50) NOT NULL,
+    registry_name VARCHAR (50) NOT NULL,
+    value TEXT NOT NULL,
+    PRIMARY KEY (key, registry_name)
```
ggarri, line 48: composed primary keys

go.mod
```diff
@@ -1,10 +1,9 @@
 module github.com/consensys/quorum-key-manager
 
-go 1.15
+go 1.16
```
ggarri, line 3: is there any special need from bumping minimum go version?

darioAnongba: reverted

deps/config/migrations/000001_create_tables.up.sql
darioAnongba: done

deps/config/migrations/000001_create_tables.up.sql
darioAnongba: done

deps/config/migrations/000001_create_tables.up.sql
darioAnongba: done

deps/config/migrations/000001_create_tables.up.sql
darioAnongba: done

deps/config/migrations/000001_create_tables.up.sql
darioAnongba: done

deps/config/migrations/000001_create_tables.down.sql
darioAnongba: done

cmd/migrate.go
darioAnongba: golang-migrate already supports SSL mode. We only need to fill the corresponding fields. I will add SSL mode in a next PR:
https://github.com/golang-migrate/migrate/blob/master/database/postgres/README.md#use-in-your-go-project

darioAnongba: implemented in next PR

ggarri (approved):

PR #278, opened 2021-08-31
Feature/259 split perms roles

### LINKED ISSUE #259: Split Roles and Permissions in JWT/TLS/API_KEY
# Description

Use a specific field for Roles and another one for Permissions

## JWT

- Use scope for Permissions
- Add property XXX for Roles

## API Key

- Use column for Permissions only
- Add columns for Roles

6l20:
## PR description

fixes #259 

## Changelog


cmd/flags/auth.go
```diff
@@ -33,6 +33,8 @@ func init() {
 	_ = viper.BindEnv(AuthOIDCClaimUsernameViperKey, authOIDCClaimUsernameEnv)
 	viper.SetDefault(AuthOIDCClaimGroupViperKey, authOIDCClaimGroupDefault)
```
ggarri, line 34: We could rename this to `AuthOIDCClaimPermissions...`

cmd/flags/auth.go
```diff
@@ -115,6 +128,7 @@ func AuthFlags(f *pflag.FlagSet) {
 	authOIDCIssuerServer(f)
 	AuthOIDCClaimUsername(f)
 	AuthOIDCClaimGroups(f)
```
ggarri, line 130: same about renaming to permissions

cmd/flags/auth.go
```diff
@@ -96,6 +102,13 @@ const (
 	authOIDCClaimGroupEnv      = "AUTH_OIDC_CLAIM_GROUPS"
 )
 
+const (
+	authOIDCClaimRolesFlag     = "auth-oidc-claim-roles"
+	authOIDCClaimRolesViperKey = "auth.oidc.claim.roles"
+	authOIDCClaimRolesDefault  = "qkm-user-roles"
```
ggarri, line 108: I prefer to use `qkm.roles` instead as it looks more standard.

cmd/flags/auth.go
```diff
@@ -285,9 +307,10 @@ func apiKeyCsvFile(vipr *viper.Viper) (map[string]apikey.UserClaims, error) {
 			return nil, fmt.Errorf("invalid number of cells in file %s should be %d", csvfile.Name(), csvRowLen)
 		}
 
-		retFile[cells[0]] = apikey.UserClaims{
-			UserName: cells[1],
-			Claims:   strings.Split(cells[2], ","),
+		retFile[cells[csvHashOffset]] = apikey.UserClaims{
+			UserName: cells[csvUserOffset],
+			Claims:   strings.Split(cells[csvClaimsOffset], ","),
```
ggarri, line 312: Claims can be renamed to Permissions now

cmd/util.go
```diff
@@ -40,10 +41,12 @@ func newUtilCommand() *cobra.Command {
 	flags.LoggerFlags(generateJWTCmd.Flags())
 	flags.AuthOIDCClaimUsername(generateJWTCmd.Flags())
 	flags.AuthOIDCClaimGroups(generateJWTCmd.Flags())
+	flags.AuthOIDCClaimRoles(generateJWTCmd.Flags())
 	flags.AuthOIDCCertKeyFile(generateJWTCmd.Flags())
 
 	generateJWTCmd.Flags().StringVar(&sub, "sub", "", "username and tenant added in claims")
-	generateJWTCmd.Flags().StringArrayVar(&scope, "scope", []string{}, "roles and permissions added in claims")
+	generateJWTCmd.Flags().StringArrayVar(&scope, "scope", []string{}, "permissions added in claims")
+	generateJWTCmd.Flags().StringArrayVar(&roles, "qkm-user-roles", []string{}, "roles added in claims")
```
ggarri, line 49: Value `"qkm-user-roles"` should be taken from flags instead of hard-coding it here

cmd/util.go
```diff
@@ -101,6 +104,7 @@ func runGenerateJWT(_ *cobra.Command, _ []string) error {
 	token, err := generator.GenerateAccessToken(map[string]interface{}{
 		oidcCfg.Claims.Subject: sub,
 		oidcCfg.Claims.Scope:   strings.Join(scope, " "),
```
ggarri, line 106: Scope renaming to Permissions

src/auth/authenticator/api-key/authenticator_test.go
```diff
@@ -24,12 +24,14 @@ const (
 
 var userAliceClaims = UserClaims{
 	UserName: "TenantOne|Alice",
-	Claims:   []string{"guest", "admin", "read:key", "write:key"},
+	Claims:   []string{"read:key", "write:key"},
```
ggarri, line 27: Renaming to permissions

src/auth/authenticator/tls/authenticator.go
```diff
@@ -52,7 +52,8 @@ func (auth Authenticator) Authenticate(req *http.Request) (*types.UserInfo, erro
 		AuthMode: AuthMode,
 	}
 	userInfo.Username, userInfo.Tenant = utils.ExtractUsernameAndTenant(clientCert.Subject.CommonName)
-	userInfo.Roles, userInfo.Permissions = utils.ExtractRolesAndPermission(clientCert.Subject.Organization)
+	userInfo.Permissions = utils.ExtractPermissions(clientCert.Subject.OrganizationalUnit)
+	userInfo.Roles = clientCert.Subject.Organization
```
ggarri, line 56: I would rather have a utility function for it

src/auth/authenticator/utils/auth.go
```diff
@@ -17,26 +19,35 @@ func ExtractUsernameAndTenant(sub string) (username, tenant string) {
...
 
-	return roles, permissions
+	return permissions
+}
+
+func ExtractClaimFromMap(claim string, claims *jwt.MapClaims) []string {
+	var roles []string
+
+	for tokenClaim, claimValue := range *claims {
+		if tokenClaim == claim {
```
ggarri, line 47: Why you couldn't do just `claims[claim]` instead?

src/auth/authenticator/oidc/authenticator.go
```diff
@@ -49,7 +49,9 @@ func (a Authenticator) Authenticate(req *http.Request) (*types.UserInfo, error)
 	}
 
 	userInfo.Username, userInfo.Tenant = utils.ExtractUsernameAndTenant(jwtData.Subject)
-	userInfo.Roles, userInfo.Permissions = utils.ExtractRolesAndPermission(jwtData.Scope)
+	userInfo.Permissions = utils.ExtractPermissions(jwtData.Scope)
+	rolesClaim := a.jwtChecker.claimsCfg.Roles
+	userInfo.Roles = utils.ExtractClaimFromMap(rolesClaim, &jwtData.MapClaims)
```
ggarri, line 54: I would rathe have the following
```
userInfo.Roles = utils.ExtractRoles(jwtData.MapClaims[roleClaim])
```

src/auth/authenticator/tls/authenticator.go
6l20: Is it necessary ? We assign an array to another one here...

ggarri: It is not very neccesary but I think it believe it would be good to have a single share method for every auth middleware to extract roles out of string or array. But this is optional

ggarri (approved):

PR #279, opened 2021-09-01
Postgres in SSL mode + migrations

darioAnongba:
Postgres in SSL mode + migrations. IMplements:

- Start postgres server in SSL mode
- Run migrations in SSL mode
- Refactor config to accept file paths
- Use file paths instead of plain values in Dialer


.env.sample
```diff
@@ -17,13 +17,9 @@ AUTH_API_KEY_FILE=/apikey/sample.csv
...
+## Postgres Configuration
+DB_TLS_SSLMODE=require
```
ggarri, line 22: Remember to update CircleCI env vars after this is merge

cmd/migrate.go
```diff
@@ -130,15 +131,26 @@ func migrateReset() error {
...
+	pgCfg, err := flags.NewPostgresConfig(vipr)
+	if err != nil {
+		errMessage := "failed to extract postgres configuration"
+		logger.WithError(err).Error(errMessage)
+		return nil, err
+	}
+
+	dbURL := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?", pgCfg.User, pgCfg.Password, pgCfg.Host, pgCfg.Port, pgCfg.Database)
+	params := url.Values{}
+	params.Add("sslmode", pgCfg.SSLMode)
```
ggarri, line 148: If SSLMode is `disabled`, does this going to work? I would suggest to set params depending on the mode being selected such as follow
```
switch pgCfg.SSLMode{
case  "verifyFull :
// Do what ever...
case "verfifyCa:
///
....
```

I think you can also get inspired by our Dialer https://github.com/ConsenSys/quorum-key-manager/blob/main/src/infra/postgres/client/dialer.go#L21

deps/docker-compose.yml
```diff
@@ -14,10 +14,17 @@ services:
     <<: *container-common
     <<: *postgres-common
     image: postgres:13.3-alpine
-    #    command: ["postgres", "-c", "max_connections=100"]
-    command: ["postgres", "-c", "log_statement=all"]
```
ggarri, line 18: could you keep this line commented as this is still useful when we are debugging?

deps/postgres/ca.crt
```diff
@@ -0,0 +1,17 @@
+-----BEGIN CERTIFICATE-----
```
ggarri, line 1: You are adding certificates twice, one under `deps/config/certificates/....` and here, I would rather keep this ones though

cmd/migrate.go
darioAnongba: Nothing should be done on our end as the package will handle everything. If you pass sslmode=disable and you also pass certificates, they will be ignored. If you pass sslmode=verify-ca without certificates it will fail etc. We don't need that switch or any other logic

deps/docker-compose.yml
darioAnongba: done

deps/postgres/ca.crt
darioAnongba: discussed. We can reuse the same client certs as http client

.env.sample
darioAnongba: yes I will update

Issue #252, opened 2021-08-09
Make QKM to perform same action on every KeyVault


ggarri:
## Acceptance criteria
- [ ] [Hashicop - secrets] Delete every secret version in  if user does not specify one
- [ ] [Hashicop - secrets] Destroy every secret version in  if user does not specify one
https://www.vaultproject.io/api/secret/kv/kv-v2#delete-metadata-and-all-versions
- [ ] Remove support for version on Delete and Destroy

PR #280, opened 2021-09-01
#252 remove version from restore,delete and destroy endpoints

ggarri:
## Fixed Issue(s)
#252 

## Changelog


src/stores/connectors/secrets/get.go
```diff
@@ -11,6 +12,14 @@ import (
 func (c Connector) Get(ctx context.Context, id, version string) (*entities.Secret, error) {
 	logger := c.logger.With("id", id, "version", version)
 
+	if version == "" {
```
darioAnongba, line 15: This should be done after the authorization check

src/stores/connectors/secrets/get.go
```diff
@@ -11,6 +12,14 @@ import (
...
+		var err error
```
darioAnongba, line 16: you don't need to define this var

src/stores/connectors/secrets/get.go
```diff
@@ -11,6 +12,14 @@ import (
...
+		version, err = c.db.GetLatestVersion(ctx, id, false)
```
darioAnongba, line 17: forgot to check that the error is not nil ? Surely missing some unit test to check this

src/stores/connectors/secrets/restore.go
```diff
@@ -9,27 +9,27 @@ import (
...
 
-	secret, err := c.db.GetDeleted(ctx, id, version)
+	secret, err := c.db.GetDeleted(ctx, id, "")
 	if err != nil {
 		return err
 	}
 
 	err = c.db.RunInTransaction(ctx, func(dbtx database.Secrets) error {
-		err = dbtx.Restore(ctx, secret.ID, version)
+		err = dbtx.Restore(ctx, secret.ID, "")
```
darioAnongba, line 27: I guess that we can also remove the `version` from the `Restore` interface?

src/stores/store/secrets/hashicorp/hashicorp.go
```diff
@@ -134,16 +138,19 @@ func (s *Store) List(_ context.Context) ([]string, error) {
...
 
-func (s *Store) Delete(_ context.Context, id, version string) error {
-	if version == "" {
-		err := errors.NotSupportedError("secret version must be specified to delete")
-		s.logger.Warn(err.Error())
-		return err
+func (s *Store) Delete(ctx context.Context, id string) error {
+	logger := s.logger.With("id", id)
+
+	versionList, err := s.db.ListVersions(ctx, id, false)
```
darioAnongba, line 144: this block can be moved to a private function given that it logs so you can return the err directly and avoid duplicating the errMessage definition. Linting should fail anyways.

src/stores/store/secrets/hashicorp/hashicorp.go
```diff
@@ -180,19 +187,19 @@ func (s *Store) ListDeleted(_ context.Context) ([]string, error) {
...
 }
 
-func (s *Store) Restore(_ context.Context, id, version string) error {
-	if version == "" {
-		err := errors.NotSupportedError("secret version must be specified to restore")
-		s.logger.Warn(err.Error())
-		return err
+func (s *Store) Restore(ctx context.Context, id string) error {
+	logger := s.logger.With("id", id)
+	versionList, err := s.db.ListVersions(ctx, id, false)
```
darioAnongba, line 192: this block can be moved to a private function given that it logs so you can return the err directly and avoid duplicating the errMessage definition. Linting should fail anyways.

src/stores/store/secrets/hashicorp/hashicorp.go
```diff
@@ -202,16 +209,19 @@ func (s *Store) Restore(_ context.Context, id, version string) error {
...
-		s.logger.Warn(err.Error())
-		return err
-	}
-
+func (s *Store) Destroy(ctx context.Context, id string) error {
 	logger := s.logger.With("id", id)
-	err := s.client.WritePost(s.pathDestroyID(id), map[string][]string{
-		"versions": {version},
+	
+	versionList, err := s.db.ListVersions(ctx, id, false)
```
darioAnongba, line 215: this block can be moved to a private function given that it logs so you can return the err directly and avoid duplicating the errMessage definition. Linting should fail anyways.

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.7.0
+	github.com/swaggo/swag v1.7.1
```
darioAnongba, line 40: this seems unexpected in this PR. Can you avoid the go.mod and go.sum changes?

src/stores/connectors/secrets/get.go
ggarri: agree

src/stores/connectors/secrets/restore.go
ggarri: I was considering too since there is not usage of no empty version but I wanted to keep the implementation. But I believe you are right, better to reduce the code to maintain than keep it for JIC reasons

go.mod
ggarri: I believe these are the lines removed by you in a previous PR because if I run `go mod tidy` + `go mod vendor` this is the current state I get. 

Do you mind to try to install swag locally using `make install-swag`  and then try again?

src/stores/connectors/secrets/get.go
ggarri: indeed I don't

src/stores/store/secrets/hashicorp/hashicorp.go
ggarri: You are right, I will encapsulate this as private method

src/stores/store/secrets/hashicorp/hashicorp.go
ggarri: You are right, I will encapsulate this as private method

darioAnongba (approved):

PR #281, opened 2021-09-02
Feature/271 idempotency

### LINKED ISSUE #271: Implement idempotency
### Summary / User Story

Implement idempotency endpoints for all possible endpoints. Normally, in a REST API, the PUT, DELETE, PATCH methods should be idempotent. 

Currently, the `restore` endpoint for example will return an error if you try to restore a key that is not deleted. This endpoint could return success if the key is already active.

6l20:
## PR description

This addresses the idempotency need on delete destroy and restore endpoints 

## Fixed Issue(s)
fixes #271 

## Changelog


src/stores/connectors/eth1/delete.go
```diff
@@ -20,25 +20,23 @@ func (c Connector) Delete(ctx context.Context, addr ethcommon.Address) error {
 	}
 
 	acc, err := c.db.Get(ctx, addr.Hex())
-	if err != nil {
-		return err
-	}
-
-	err = c.db.RunInTransaction(ctx, func(dbtx database.ETH1Accounts) error {
-		err = dbtx.Delete(ctx, addr.Hex())
+	if err == nil {
```
ggarri, line 23: According to what it is described at the rest-api doc https://restfulapi.net/idempotent-rest-apis/ 
```
HTTP DELETE
When you invoke N similar DELETE requests, first request will delete the resource and response will be 200 (OK) or 204 (No Content). Other N-1 requests will return 404 (Not Found). Clearly, the response is different from first request, but there is no change of state for any resource on server side because original resource is already deleted. So, DELETE is idempotent.
```

it should fail with 404 when the asset is already deleted, therefore previous behaviour was okay.

src/stores/connectors/eth1/destroy.go
```diff
@@ -19,26 +19,30 @@ func (c Connector) Destroy(ctx context.Context, addr ethcommon.Address) error {
...
+	_, err = c.Get(ctx, addr)
+	if err == nil {
+		logger.WithError(err).Error("try to destroy an exiting eth1Account, must be deleted first")
+		return errors.StatusConflictError("account %s must be deleted first", addr.Hex())
 	}
 
-	err = c.db.RunInTransaction(ctx, func(dbtx database.ETH1Accounts) error {
-		err = dbtx.Purge(ctx, addr.Hex())
+	acc, err := c.db.GetDeleted(ctx, addr.Hex())
+	if err == nil {
```
ggarri, line 29: Previous behaviour was okay 
```
HTTP DELETE
When you invoke N similar DELETE requests, first request will delete the resource and response will be 200 (OK) or 204 (No Content). Other N-1 requests will return 404 (Not Found). Clearly, the response is different from first request, but there is no change of state for any resource on server side because original resource is already deleted. So, DELETE is idempotent.
```

src/stores/connectors/eth1/restore.go
```diff
@@ -19,26 +19,29 @@ func (c Connector) Restore(ctx context.Context, addr ethcommon.Address) error {
...
-	if err != nil {
-		return err
+	_, err = c.Get(ctx, addr)
+	if err == nil {
+		return nil
 	}
 
-	err = c.db.RunInTransaction(ctx, func(dbtx database.ETH1Accounts) error {
-		err = dbtx.Restore(ctx, addr.Hex())
+	acc, err := c.db.GetDeleted(ctx, addr.Hex())
```
ggarri, line 27: If key exists already, should so return as you do in line 22-24, but rest of code should have remained the same way as it was

src/stores/connectors/eth1/delete.go
6l20: Yes right, I did roll back

darioAnongba (approved):

PR #282, opened 2021-09-02
Generate valid certificates using CA

darioAnongba:
Generate valid certificates using CA


.env.sample
```diff
@@ -1,25 +1,7 @@
 LOG_LEVEL=debug
 LOG_FORMAT=text
 
-HTTP_PORT=8080
```
ggarri, line 4: IMO it is better to keep all setting in the ENV instead of modifying the docker-composer file.  Otherwise it wouldn't be possible to run qkm only with OIDC or without TLS etc....

docker-compose.yml
```diff
@@ -5,13 +5,13 @@ x-default-variables: &default-variables
...
-  HTTP_PORT: ${HTTP_PORT-}
+  HTTP_PORT: ${HTTP_PORT-8080}
   HTTP_HOST: ${HTTP_HOST-0.0.0.0}
-  HEALTH_PORT: ${HEALTH_PORT-}
-  DB_TLS_SSLMODE: ${DB_TLS_SSLMODE-}
-  DB_TLS_CERT: ${DB_TLS_CERT-}
-  DB_TLS_KEY: ${DB_TLS_KEY-}
-  DB_TLS_CA: ${DB_TLS_CA-}
+  HEALTH_PORT: ${HEALTH_PORT-8081}
+  DB_TLS_SSLMODE: ${DB_TLS_SSLMODE-verify-ca}
```
ggarri, line 11: As we did in Orchestrate, don't set values in docker compose but in .env so we can make custome settings with it, otherwise for circleCI I believe it would be hard to know what is default and by is overwritten etc

ggarri (approved):

PR #283, opened 2021-09-03
Fix e2e tests

darioAnongba:
PR to fix e2e tests by generating new certificates and using the required ones


deps/config/apikey/sample.csv
```diff
@@ -4,4 +4,4 @@
 # sha256("admin-user")
 f470213d9ae659187a19b9cb2169b4b400544f4d3f59250eda657154700da616;tenant1:user1;*:*;admin
 # sha256("guest-user")
-8605f70ff5f55e2a9323d97de3dbf8e61f38314d93298ca00e19f8918fe8971b;tenant2:user2;*:secrets,*:keys,read:eth1accounts;signer
+8605f70ff5f55e2a9323d97de3dbf8e61f38314d93298ca00e19f8918fe8971b;tenant2:user2;*:secrets,*:keys,read:eth1accounts;
```
6l20, line 7: Last column should be the role here, was it removed in code ?

deps/ca/ca.srl
```diff
@@ -1 +1 @@
-554E0AF33CB76CA1A5B493A14C2F1170C290B4B2
+554E0AF33CB76CA1A5B493A14C2F1170C290B4B4
```
6l20, line 1: Do we really need that srl file here ?

deps/config/apikey/sample.csv
darioAnongba: no role is needed for "guest-user" , this breaks the e2e tests.

deps/ca/ca.srl
darioAnongba: removed

6l20 (approved):

PR #286, opened 2021-09-05
fix local keys error with using Hashicorp

ggarri:
## PR description
- Modified LocalKeys to persist secrets in DB so that we can list versions for Restoring and Destroying in Hashicorp

## Fixed Issue(s)
- Fixed error on Restore and Destroy with LocalKeys + Hashicorp 

## Changelog

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -14,9 +14,9 @@ AUTH_API_KEY_FILE=/apikey/sample.csv
 
 TEST_DATA={"key_manager_url":"https://localhost:8080","health_key_manager_url":"https://localhost:8081","secret_stores":["hashicorp-secrets","akv-secrets","aws-secrets"],"key_stores":["hashicorp-keys"],"eth1_stores":["eth1-accounts"],"quorum_node_id":"quorum-node","besu_node_id":"besu-node"}
 
-## Postgres Configuration
-DB_TLS_SSLMODE=verify-ca
-DB_TLS_CERT=/certificates/client.crt
-DB_TLS_KEY=/certificates/client.key
-DB_TLS_CA=/ca/ca.crt
-DB_HOST=postgres-ssl
+## Postgres SSL
+#DB_TLS_SSLMODE=verify-ca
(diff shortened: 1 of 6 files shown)


.env.sample
```diff
@@ -14,9 +14,9 @@ AUTH_API_KEY_FILE=/apikey/sample.csv
 
 TEST_DATA={"key_manager_url":"https://localhost:8080","health_key_manager_url":"https://localhost:8081","secret_stores":["hashicorp-secrets","akv-secrets","aws-secrets"],"key_stores":["hashicorp-keys"],"eth1_stores":["eth1-accounts"],"quorum_node_id":"quorum-node","besu_node_id":"besu-node"}
 
-## Postgres Configuration
-DB_TLS_SSLMODE=verify-ca
-DB_TLS_CERT=/certificates/client.crt
-DB_TLS_KEY=/certificates/client.key
-DB_TLS_CA=/ca/ca.crt
-DB_HOST=postgres-ssl
+## Postgres SSL
```
darioAnongba, line 17: this is no longer working?

src/stores/manager/keys/local.go
```diff
@@ -55,5 +55,5 @@ func NewLocalKeyStore(specs *LocalKeySpecs, db database.Secrets, logger log.Logg
 		return nil, err
 	}
 
-	return localkeys.New(secretStore, logger), nil
+	return localkeys.New(secretStore, db, logger), nil
```
darioAnongba, line 58: I don't think this is how this should be implemented. IMO the local keys should accept a secret connector as dependency instead of a secret store. The secret connector already has all the functions you implemented

darioAnongba (approved):

.env.sample
ggarri: I commented for our dev envs, since I was running into the permission issue of the certs. IMO it is not very necessary to be running local dev with SSL PG if that brings complexity to our dev environments. I would rather validate the SSL only on our e2e once we get to run it using remote PG

PR #290, opened 2021-09-06
Renaming ethereum endpoints

darioAnongba:
Rename `eth` to `ethereum` in endpoints and use `eth` instead of `eth1` internally as naming


dolanor:
Why the eth1 name was selected? In case we would support eth2 as well? Is there a plan to have some eth2 quorum?

deps/ca/ca.srl
```diff
@@ -0,0 +1 @@
+20EBCB11194D5794BB66DBC343B86D9D15BA47EB
```
ggarri, line 1: I thought you decied on removing them. Why did you add it again?

deps/config/apikey/sample.csv
```diff
@@ -2,6 +2,6 @@
 ## Column separator is ";"
 
 # sha256("admin-user")
-f470213d9ae659187a19b9cb2169b4b400544f4d3f59250eda657154700da616;tenant1:user1;*:*;admin
+f470213d9ae659187a19b9cb2169b4b400544f4d3f59250eda657154700da616;tenant1|user1;*:*;admin
 # sha256("guest-user")
-8605f70ff5f55e2a9323d97de3dbf8e61f38314d93298ca00e19f8918fe8971b;tenant2:user2;*:secrets,*:keys,read:eth1accounts;
+8605f70ff5f55e2a9323d97de3dbf8e61f38314d93298ca00e19f8918fe8971b;tenant2|user2;*:secrets,*:keys,read:ethaccounts;anonymous
```
ggarri, line 7: What was the final resource name? `ethereum` or `ethaccount` ? In the template you are using `ethereum` and here `ethaccount`

go.mod
```diff
@@ -30,14 +30,14 @@ require (
...
 	github.com/justinas/alice v1.2.0
+	github.com/lib/pq v1.8.0
 	github.com/magefile/mage v1.10.0 // indirect
 	github.com/oxtoacart/bpool v0.0.0-20190530202638-03653db5a59c
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.7.0
-	github.com/swaggo/swag v1.7.1
```
ggarri, line 40: We keep going back and forth with this? this you try to run `$> make install-swag` already?

deps/ca/ca.srl
darioAnongba: mistake

deps/config/manifests/roles.yml.template
```diff
@@ -14,7 +14,7 @@
     permission:
       - "read:*"
       - "sign:keys"
-      - "sign:eth1accounts"
+      - "sign:ethereum"
```
ggarri, line 17: I think this should be `sign:ethaccount`

src/auth/authenticator/tls/authenticator_test.go
```diff
@@ -57,7 +57,7 @@ func TestAuthenticatorSameCert(t *testing.T) {
 		assert.Equal(t, "eve", userInfo.Username)
 		assert.Equal(t, "auth0", userInfo.Tenant)
 		assert.Equal(t, []string{"signer"}, userInfo.Roles)
-		assert.Equal(t, types.ListWildcardPermission("*:eth1accounts"), userInfo.Permissions)
+		assert.Equal(t, types.ListWildcardPermission("*:ethereum"), userInfo.Permissions)
```
ggarri, line 60: Again, I am confused. Should it be `ethaccounts` ?

deps/ca/ca.srl
darioAnongba: removed

deps/config/apikey/sample.csv
darioAnongba: mistake

darioAnongba: changed

go.mod
darioAnongba: Is it okay now? I ran the command and then `go mod tidy`

src/auth/authenticator/tls/authenticator_test.go
darioAnongba: why? `ethereum` is clearly a better name than `ethaccounts` since the domain is named `ethereum` so it correlates with keys and secrets

deps/config/manifests/roles.yml.template
darioAnongba: why? `ethereum` is clearly a better name than `ethaccounts` since the domain is named `ethereum` so it correlates with keys and secrets

src/auth/manager/service_test.go
```diff
@@ -32,19 +32,19 @@ var testManifest = []byte(`
   name: signer
   specs:
     permission:
-      - read:eth1accounts
+      - read:ethaccounts
```
ggarri, line 35: I guess this should be renamed to `read:ethereum`, right?

deps/config/manifests/store.yml.template
```diff
@@ -58,9 +58,9 @@
       address: http://hashicorp:8200
       token: '{VAULT_TOKEN}'
       namespace: ''
-- kind: Eth1Account
+- kind: EthAccount
```
ggarri, line 61: I would rather rename this to "Ethereum" to correlate with the rest of the changes

src/auth/manager/service_test.go
darioAnongba: lol why are these tests passing. Updated

deps/config/manifests/store.yml.template
darioAnongba: applied

ggarri (approved):

PR #291, opened 2021-09-07
added unauthorized and forbidden docs

ggarri:
## Changelog


6l20 (approved):

dolanor (approved):
```
// @Failure 401 {object} ErrorResponse "Unauthorized"
// @Failure 403 {object} ErrorResponse "Forbidden"
```
We know 401 is unauthorized and 403 is forbidden by definition in HTTP, adding it as a comment does add much value.
But other than that, LGTM

PR #294, opened 2021-09-07
36 pagination

ggarri:
## PR description
- Include pagination arguments on list endpoints for secrets, keys and ethereum accounts

## Fixed Issue(s)
#39 

## Changelog


src/stores/api/handlers/stores.go
```diff
@@ -46,8 +51,42 @@ func (h *StoresHandler) Register(router *mux.Router) {
...
 }
 
-func StoreSelector(h http.Handler) http.Handler {
+func storeSelector(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		h.ServeHTTP(w, r.WithContext(WithStoreName(r.Context(), mux.Vars(r)["storeName"])))
 	})
 }
+
+func getLimitOffset(request *http.Request) (int, int, error) {
```
darioAnongba, line 60: something is weird with the logic of this function. Shouldn't the logic be:
- `limit` and `offset` are DEFAULT_PAGE_SIZE, 0 by default (named vars from signature)
- If `rLimit` and `rPage` are empty: return DEFAULT_PAGE_SIZE, 0, nil
- If `rLimit` is not empty: limit = transform to uint (no negative values) rLimit
- If `rPage` is not empty: offset = limit * transform to uint(rOffset)
- return limit, offset

Basically reversing the logic.

src/stores/api/handlers/stores.go
```diff
@@ -46,8 +51,42 @@ func (h *StoresHandler) Register(router *mux.Router) {
...
+		limit = DEFAULT_PAGE_SIZE
+	}
+
+	if limit == "" {
+		return 0, 0, nil
+	}
+
+	iLimit, err := strconv.Atoi(limit)
+	if err != nil {
+		return 0, 0, errors.InvalidParameterError("invalid limit value")
```
darioAnongba, line 77: InvalidFormat for 400 here

src/stores/api/handlers/stores.go
```diff
@@ -46,8 +51,42 @@ func (h *StoresHandler) Register(router *mux.Router) {
...
+	}
+
+	iPage := 0
+	iOffset := 0
+	if page != "" {
+		iPage, err = strconv.Atoi(page)
+		if err != nil {
+			return 0, 0, errors.InvalidParameterError("invalid page value")
```
darioAnongba, line 85: InvalidFormat for 400 here

src/stores/api/handlers/stores.go
darioAnongba: better name the returned values since the function returns 3

src/stores/connectors/ethereum/list.go
```diff
@@ -8,40 +8,40 @@ import (
...
 
-func (c Connector) ListDeleted(ctx context.Context) ([]common.Address, error) {
+func (c Connector) ListDeleted(ctx context.Context, limit, offset int) ([]common.Address, error) {
 	err := c.authorizator.CheckPermission(&types.Operation{Action: types.ActionRead, Resource: types.ResourceEthAccount})
 	if err != nil {
 		return nil, err
 	}
 
-	accs, err := c.db.GetAllDeleted(ctx)
+	strAddr, err := c.db.ListAddresses(ctx, true, limit, offset)
```
darioAnongba, line 37: Can we rename `ListAddresses` to `SearchAddresses` ? like we have in Orchestrate. Seems better to me.

src/stores/connectors/keys/list.go
```diff
@@ -6,42 +6,32 @@ import (
...
-func (c Connector) ListDeleted(ctx context.Context) ([]string, error) {
+func (c Connector) ListDeleted(ctx context.Context, limit, offset int) ([]string, error) {
 	err := c.authorizator.CheckPermission(&types.Operation{Action: types.ActionRead, Resource: types.ResourceKey})
 	if err != nil {
 		return nil, err
 	}
 
-	ids := []string{}
-	keysRetrieved, err := c.db.GetAllDeleted(ctx)
+	ids, err := c.db.ListIDs(ctx, true, limit, offset)
```
darioAnongba, line 30: Same for SearchIDs

src/stores/database/postgres/ethereum.go
```diff
@@ -99,6 +100,35 @@ func (ea *ETHAccounts) GetAllDeleted(ctx context.Context) ([]*entities.ETHAccoun
 	return accounts, nil
 }
 
+func (s *ETHAccounts) ListAddresses(ctx context.Context, isDeleted bool, limit, offset int) ([]string, error) {
```
darioAnongba, line 103: could this return []common.Address ?

src/stores/database/postgres/ethereum.go
```diff
@@ -99,6 +100,35 @@ func (ea *ETHAccounts) GetAllDeleted(ctx context.Context) ([]*entities.ETHAccoun
...
+	var ids = []string{}
+	var err error
+	var query string
+	args := []interface{}{s.storeID}
+
+	switch {
+	case limit != 0 || offset != 0:
```
darioAnongba, line 110: can we have a generic function in `infra` to do this ? Like one accepting the id field name (address or id), limit and offset and returns a list of string? The function looks very similar in all 3 agents

src/stores/database/postgres/ethereum.go
ggarri: I also considered that, but I decided on this so we dont not tight out Data layer to anything else but Entities. Let me know if you think otherwise

src/stores/api/handlers/stores.go
ggarri: > limit and offset are DEFAULT_PAGE_SIZE, 0 by default (named vars from signature)
Nop

> If rLimit and rPage are empty: return DEFAULT_PAGE_SIZE, 0, nil
I don't want to force any limit size without user's to now, so that is why I implemented in a way there is not limit size UNLESS user set a page value

>If rLimit is not empty: limit = transform to uint (no negative values) rLimit
good point

>If rPage is not empty: offset = limit * transform to uint(rOffset)
Done

> return limit, offset
I still need to check for errors on values

src/stores/api/handlers/stores.go
ggarri: Indeed

src/stores/api/handlers/stores.go
ggarri: done

src/stores/connectors/ethereum/list.go
ggarri: But we are not doing any search, we are actually just listing. In Orchestrate we named "Search" because it was an actual search method which return a list of items, in this case we are listing a ids/addresses

src/stores/connectors/keys/list.go
ggarri: same comment as above

src/stores/database/postgres/ethereum.go
ggarri: I also considered but the complexity of the method it would be too high for not much value as we alredy know this won't be extended further than this 3 agents. Although happy to refactor if you think otherwise on this

src/stores/connectors/ethereum/list.go
darioAnongba: But I believe we will use the same function to Search if we have to in the future. Searching will return a list of IDs in the QKM. 

We will pass filters to this function to achieve this. Limit and offset are examples of filters already and you are searching for specific IDs.

src/stores/database/postgres/ethereum.go
darioAnongba: But common.Address is an entity right? It's just not one of our entities. But if you don't want to be tied to an external package it's fine to me.

src/stores/database/postgres/ethereum.go
darioAnongba: It could be extended to aliases, roles, nodes and stores. You can list those by id as well even if it's unlikely at the moment that we implement this.

src/stores/connectors/ethereum/list.go
ggarri: I see your point, but I barely see this function will evolve that way. Lets modify it though and let see :)

src/stores/database/postgres/ethereum.go
ggarri: Okay, done

src/infra/postgres/client/queries.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func QuerySearchIDs(ctx context.Context, client postgres.Client, idCol, whereCond string, whereArgs []interface{}, isDeleted bool, limit, offset uint64) ([]string, error) {
+	var ids = []string{}
+	var err error
+	var query string
+
+	switch {
+	case limit != 0 || offset != 0:
+		query = fmt.Sprintf("SELECT (array_agg(%s ORDER BY created_at ASC))[%d:%d] FROM secrets WHERE %s", idCol, offset+1, offset+limit, whereCond)
```
darioAnongba, line 17: How are your acceptance tests passing if the table is hardcoded to `secrets` here ? :thinking:

ggarri: becuase it was not passing xD. Good catch!

darioAnongba (approved):

Issue #148, opened 2021-06-15
Allow to pass alias placeholders in JSON-RPC request


nmvalera:
#### Summary

Allow to next arguments pass JSON-RPC to replace
- `privateFor` c.f. https://docs.google.com/document/d/1cgfbBb7EJ-CAvkR21k0h4GdEC_vS1XdMrEg-PIPvaak/edit#heading=h.hebsvnkcovnm
- `privacyGroupId` c.f. https://docs.google.com/document/d/1cgfbBb7EJ-CAvkR21k0h4GdEC_vS1XdMrEg-PIPvaak/edit#heading=h.njkpsii4nbvo

#### Acceptance Criteria

- [x] Enable to parse alias placeholders in JSON-RPC request (the current proposal is `{{...}}`)
- [x] Allow `eth_sendTransaction` interceptor to use alias in `privateFor`
- [x] #377
- [x] #393
- [x] #388
- [x] Allow `eth_sendRawPrivateTransaction` interceptor to use alias in `privacyGroupId` argument
- [x] Allow `eea_sendTransaction` interceptor to use alias in  `privateFor`
- [x] Allow `eea_sendTransaction` interceptor to use alias in  `privacyGroupId` (in this case if the alias value is a string it should be treated as the privacy group id, if alias value is a string array then it should be treated as an array to pass in `privateFor`
- [x] Allow `eea_sendTransaction` interceptor to use alias in `privateFrom`
- [x] E2E tests for all JSON-RPC methods

dolanor:
All the work on EEA has been merged with #421 
There is one test missing in the E2E, it is the test of the PrivacyGroupID alias pointing to a PrivacyGroupID (as a string), as opposed as a PrivacyGroupID alias pointing to an array alias. It would be good to add it to ensure completeness.

PR #295, opened 2021-09-07
parse aliases in addresses for eth_sendTransaction

dolanor:
## PR description

First step to take care of #148


src/aliases/manager/manager.go
```diff
@@ -0,0 +1,52 @@
...
+// Start does nothing as the DB client is already connected.
+func (m *BaseManager) Start(_ context.Context) error { return nil }
+
+// Stop does nothing as the DB client should be stopped outside of it.
+func (m *BaseManager) Stop(_ context.Context) error { return nil }
+func (m *BaseManager) Error() error                 { return nil }
+func (m *BaseManager) Close() error                 { return nil }
+
+func (m *BaseManager) CreateAlias(ctx context.Context, registry aliasent.RegistryName, alias aliasent.Alias) (*aliasent.Alias, error) {
+	return m.db.Alias().CreateAlias(ctx, registry, alias)
```
darioAnongba, line 31: Any reason why your use cases are not doing anything? You probably need permission checks and logging here. (same for all use cases)

src/aliases/store/models/alias.go
```diff
@@ -15,29 +22,64 @@ type Alias struct {
...
+		if err != nil {
+			return errors.InvalidFormatError("alias value is incorrect: %v: %v", err, v)
+		}
+	}
+	return nil
+}
+
+func AliasFromEntity(ent aliasent.Alias) (alias Alias, err error) {
+	// check that the ent.Value only contains base64 string
+	err = checkAliasValue(ent.Value)
```
darioAnongba, line 38: It is not the responsibility of this function (or the data layer) to do this check. This check is business logic. Most of the time (if not always), constructors such as this should not return an error and `InvalidParameterErrors` come from the business layer.

src/aliases/store/models/alias.go
```diff
@@ -15,29 +22,64 @@ type Alias struct {
...
+	if err != nil {
+		return alias, err
+	}
+
+	var b bytes.Buffer
+	err = json.NewEncoder(&b).Encode(ent.Value)
```
darioAnongba, line 44: if I understand correctly, here you are transforming `value` into its JSON representation. If that's correct then this operation here is unnecessary.

src/aliases/store/models/alias.go
```diff
@@ -15,29 +22,64 @@ type Alias struct {
...
 
-func (a *Alias) ToEntity() *aliasent.Alias {
+func (a *Alias) ToEntity() (*aliasent.Alias, error) {
+	var value aliasent.AliasValue
+	err := json.Unmarshal([]byte(a.Value), &value)
+	if err != nil {
+		return nil, err
+	}
+
+	err = checkAliasValue(value)
```
darioAnongba, line 62: Unnecessary check here. You are checking that data coming from the storage is properly formatted but that is an assertion of the app (not an exception).

src/aliases/store/models/alias.go
```diff
@@ -15,29 +22,64 @@ type Alias struct {
...
-		Value:        AliasValue(ent.Value),
-	}
+		Value:        AliasValue(b.String()),
+	}, nil
 }
 
-func (a *Alias) ToEntity() *aliasent.Alias {
+func (a *Alias) ToEntity() (*aliasent.Alias, error) {
+	var value aliasent.AliasValue
+	err := json.Unmarshal([]byte(a.Value), &value)
```
darioAnongba, line 57: Unnecessary unmarshalling

src/aliases/store/models/alias.go
```diff
@@ -15,29 +22,64 @@ type Alias struct {
...
 		Key:          aliasent.AliasKey(a.Key),
 		RegistryName: aliasent.RegistryName(a.RegistryName),
-		Value:        aliasent.AliasValue(a.Value),
-	}
+		Value:        value,
+	}, nil
 }
 
-func AliasesToEntity(aliases []Alias) []aliasent.Alias {
+func AliasesToEntity(aliases []Alias) ([]aliasent.Alias, error) {
```
darioAnongba, line 73: This function does not belong here. It seems to be a utility function of a specific use case.

src/aliases/store/postgres/alias.go
```diff
@@ -23,10 +23,13 @@ func NewAlias(pgClient postgres.Client) *AliasStore {
 }
 
 func (s *AliasStore) CreateAlias(ctx context.Context, registry aliasent.RegistryName, alias aliasent.Alias) (*aliasent.Alias, error) {
-	a := aliasmodels.AliasFromEntity(alias)
+	a, err := aliasmodels.AliasFromEntity(alias)
```
darioAnongba, line 26: This function should not return an error at this level.  Same comment for all functions of this Agent/Repository

src/aliases/store/postgres/alias.go
```diff
@@ -23,10 +23,13 @@ func NewAlias(pgClient postgres.Client) *AliasStore {
...
+	if err != nil {
+		return nil, err
+	}
 	a.RegistryName = aliasmodels.RegistryName(registry)
 
-	err := s.pgClient.Insert(ctx, &a)
+	err = s.pgClient.Insert(ctx, &a)
```
darioAnongba, line 32: You should be logging here and returning a specific error message. Same comment for all functions of this Agent/Repository. Otherwise you are returning specific postgres errors to the users.

src/nodes/interceptor/eth_send_transaction.go
```diff
@@ -54,6 +56,13 @@ func (i *Interceptor) sendPrivateTx(ctx context.Context, msg *ethereum.SendTxMsg
 		msg.Data = new([]byte)
 	}
 
+	// extract aliases from PrivateFor
+	*msg.PrivateFor, err = placeholder.ReplaceAliases(ctx, i.aliases, *msg.PrivateFor)
+	if err != nil {
+		i.logger.WithError(err).Error("failed to replace aliases in privateFor")
+		return nil, errors.BlockchainNodeError(err.Error())
```
darioAnongba, line 63: `BlockchainNodeError` doesn't seem to be the correct error here.

src/aliases/placeholder/parser.go
```diff
@@ -0,0 +1,60 @@
...
+	"regexp"
+	"sync"
+
+	aliasent "github.com/consensys/quorum-key-manager/src/aliases/entities"
+)
+
+const aliasParseFormat = `{{(?m)(?P<registry>[a-zA-Z0-9-_+]+):(?P<alias>[a-zA-Z0-9-_+]+)}}$`
+
+var (
+	aliasParseRegexOnce sync.Once
```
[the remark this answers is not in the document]
dolanor: It's a common way to do in Go for that specific case. The fact that it is a global is mellowed by the fact it is scoped by the package, and mostly it is private.
Why I did that is because the initialization of a regexp pattern is expensive and should be done once.
In general, I'm all against global var, but adding a specific struct to just init seems it's gonna add some complexity.
But in another sense, it does contain business logic.

src/aliases/manager/manager.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Start does nothing as the DB client is already connected.
+func (m *BaseManager) Start(_ context.Context) error { return nil }
+
+// Stop does nothing as the DB client should be stopped outside of it.
+func (m *BaseManager) Stop(_ context.Context) error { return nil }
+func (m *BaseManager) Error() error                 { return nil }
+func (m *BaseManager) Close() error                 { return nil }
+
+func (m *BaseManager) CreateAlias(ctx context.Context, registry aliasent.RegistryName, alias aliasent.Alias) (*aliasent.Alias, error) {
```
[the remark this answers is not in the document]
dolanor: Are you sure about the `alias` sub-folder as well? `src/aliases/{folder_name}/alias/`?
Personally, I think this is part of the discussion we need to have on how we organize the project/component/sub-folders.
I suggest we keep the refactoring for later?

src/aliases/manager/manager.go
dolanor: There were no permissions that I knew of when I started. I will add this as an issue.
What's the verbosity of the logging we want? We enter the func, we log it? We log only on error?

I just created a boilerplate that is accessible from the stores service.

src/aliases/store/models/alias.go
dolanor: How do make sure the data is correctly formatted/coherent without error before entering the data storage?

src/aliases/store/models/alias.go
dolanor: How come?

src/aliases/store/models/alias.go
dolanor: How come?

src/aliases/store/models/alias.go
dolanor: ok

src/aliases/store/postgres/alias.go
dolanor: So what should we do about the error that could be raised inside it?

src/aliases/store/postgres/alias.go
dolanor: For me logging is already dealing with the error. Logging + returning an error could be doing the work twice.
We need to add that to the list :)

src/nodes/interceptor/eth_send_transaction.go
dolanor: You're right, I think it's a copypasta…

dolanor: fixed in dca55e7

src/aliases/placeholder/parser.go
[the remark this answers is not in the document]
dolanor: done in cf1d1e1

src/aliases/manager/manager.go
dolanor: We need to create an issue about the authorizations for alias creation/access. The implications are big (rewriting the destination of an alias for another address is really dangerous)
#345

src/aliases/store/models/alias.go
dolanor: done

src/aliases/store/models/alias.go
dolanor: go-pg driver automatically marshall Go struct to its JSON representation if the destination is a TEXT/JSON in the DB.

I removed it.

src/aliases/store/models/alias.go
dolanor: go-pg driver automatically marshall Go struct to its JSON representation if the destination is a TEXT/JSON in the DB.

I removed it.

src/aliases/store/models/alias.go
dolanor: It's a utility func that has its place here as it only manipulates the models types from/to entities.

src/aliases/store/postgres/alias.go
dolanor: I removed the error from the func as we don't need to marshal to JSON anymore.

src/aliases/api/handlers/handlers.go
```diff
@@ -295,3 +306,14 @@ func validatePathVars(pathVars ...string) error {
 	}
 	return nil
 }
+
+func checkAliasValue(value types.AliasValue) error {
+	for _, v := range value {
+		// we just want the decoding to happen without error
+		_, err := base64.StdEncoding.DecodeString(v)
+		if err != nil {
+			return errors.InvalidFormatError("alias value is incorrect: %v: %v", err, v)
```
darioAnongba, line 315: this function should either return `InvalidFormatError` and then the parent should just return it or this function should return a normal error and the parent wraps it in an `InvalidFormatError`. Currently there is a double return of `InvalidFormatError` from parents and this function.

src/aliases/api/handlers/handlers_test.go
```diff
@@ -222,7 +222,7 @@ func TestGetAlias(t *testing.T) {
 		c := defaultCase()
 		c.key = nonExistingKey
 		c.status = http.StatusNotFound
-		ent := newEntAlias(c.reg, c.key, "")
+		ent := newEntAlias(c.reg, c.key, nil)
```
darioAnongba, line 225: is this a faker function? In this case it should be placed in a folder called `testutils` and be called `FakeAlias`. Check the other testutils from `stores` domain for example.

src/aliases/api/types/alias.go
```diff
@@ -34,7 +34,7 @@ func FormatEntityAliases(ents []aliasent.Alias) []Alias {
 	return als
 }
 
-type AliasValue string
+type AliasValue []string
 
 type AliasKey string
```
darioAnongba, line 39: minor: I still personally think that this habit of redeclaring types in Go is a bit overkill as using string seems fine.

src/aliases/connectors/aliases/aliases.go
```diff
@@ -0,0 +1,77 @@
...
+func (m *Connector) UpdateAlias(ctx context.Context, registry aliasent.RegistryName, alias aliasent.Alias) (*aliasent.Alias, error) {
+	logger := m.logger.With(
+		"registry_name", registry,
+		"alias_key", alias.Key,
+	)
+	a, err := m.db.Alias().UpdateAlias(ctx, registry, alias)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println("alias updated successfully")
```
darioAnongba, line 53: forgot println

src/aliases/connectors/aliases/aliases.go
```diff
@@ -0,0 +1,77 @@
...
+	logger.Info("alias deleted successfully")
+	return nil
+}
+
+func (m *Connector) ListAliases(ctx context.Context, registry aliasent.RegistryName) ([]aliasent.Alias, error) {
+	return m.db.Alias().ListAliases(ctx, registry)
+}
+
+func (m *Connector) DeleteRegistry(ctx context.Context, registry aliasent.RegistryName) error {
+	return m.db.Alias().DeleteRegistry(ctx, registry)
```
darioAnongba, line 76: forgot logging

src/aliases/placeholder/parser.go
```diff
@@ -0,0 +1,62 @@
...
+
+type Parser struct {
+	regex *regexp.Regexp
+}
+
+func New() (*Parser, error) {
+	const aliasParseFormat = `{{(?m)(?P<registry>[a-zA-Z0-9-_+]+):(?P<alias>[a-zA-Z0-9-_+]+)}}$`
+	regex, err := regexp.Compile(aliasParseFormat)
+	if err != nil {
+		return nil, errors.InvalidFormatError("bad regexp format '%v': %v", aliasParseFormat, err)
```
darioAnongba, line 19: this seems to be a `ConfigError` instead as the regex is part of the codebase.

src/aliases/placeholder/parser.go
```diff
@@ -0,0 +1,62 @@
...
+	aliasent "github.com/consensys/quorum-key-manager/src/aliases/entities"
+)
+
+type Parser struct {
+	regex *regexp.Regexp
+}
+
+func New() (*Parser, error) {
+	const aliasParseFormat = `{{(?m)(?P<registry>[a-zA-Z0-9-_+]+):(?P<alias>[a-zA-Z0-9-_+]+)}}$`
+	regex, err := regexp.Compile(aliasParseFormat)
```
darioAnongba, line 17: good stuff

darioAnongba (approved):

src/aliases/api/handlers/handlers.go
dolanor: good catch

src/aliases/api/types/alias.go
dolanor: Yeah, I think I overdid it as well, it was to validate the strings potentially, but as those data are short lived and validation is done by go-playground, having a dedicated type is overkill.

src/aliases/api/handlers/handlers_test.go
dolanor: I will put it in testutils, but I believe it would make more sense in a `testdata` folder which will make any public faker functions available from tests written in any package, without leaking it in the rest of the application. I will add this nuance in the architecture doc.

src/aliases/connectors/aliases/aliases.go
dolanor: done

src/aliases/connectors/aliases/aliases.go
dolanor: done

src/aliases/placeholder/parser.go
dolanor: Yes, I guess I couldn't see which type corresponded best. We should document what error does what.

src/aliases/api/types/alias.go
dolanor: done

dolanor:
Normally I fixed everything and made the file/folder structure mirror of that from `store`

src/aliases/api/handlers/handlers.go
```diff
@@ -173,33 +160,32 @@ func (h *AliasHandler) updateAlias(w http.ResponseWriter, r *http.Request) {
...
+	var aliasReq types.AliasRequest
+	err := jsonutils.UnmarshalBody(r.Body, &aliasReq)
 	if err != nil {
 		infrahttp.WriteHTTPErrorResponse(w, errors.InvalidFormatError(err.Error()))
 		return
 	}
-
-	var aliasReq types.AliasRequest
-	err = jsonutils.UnmarshalBody(r.Body, &aliasReq)
+	err = checkAliasValue(aliasReq.Value)
```
darioAnongba, line 169: This function seems unnecessary, why is it not in the JSON annotation?

src/aliases/api/types/alias.go
```diff
@@ -34,16 +37,12 @@ func FormatEntityAliases(ents []aliasent.Alias) []Alias {
...
-type AliasValue string
-
-type AliasKey string
-
-type RegistryName string
-
+// AliasRequest creates or modifies an alias value.
 type AliasRequest struct {
-	Value AliasValue `json:"value"`
+	Value []string `json:"value"`
```
darioAnongba, line 42: forgot the rest of the annotations, check how it's done in the other domains but you need a validate and an example annotation.

src/aliases/parser/alias_parser.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+
+	"github.com/consensys/quorum-key-manager/src/aliases"
+)
+
+//go:generate mockgen -source=alias_parser.go -destination=mock/alias_parser.go -package=mock
+
+// AliasParser parses and replace aliases.
+type AliasParser interface {
```
darioAnongba, line 12: To be honest I liked better your proposal to completely remove this and to add these use cases to the connectors. IMO you just created this component because of the way Nico created the tickets: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/148 `(Allow to pass alias placeholders)`

That's why you separated it in a different component and called it "placeholder", but fundamentally this just goes into your AliasBackend.

src/aliases/parser/parser.go
```diff
@@ -0,0 +1,67 @@
...
+
+	regName = submatches[1]
+	aliasKey = submatches[2]
+
+	return regName, aliasKey, true
+}
+
+// ReplaceAliases replace a slice of potential aliases with a slice having all the aliases replaced by their value.
+// It will fail if no aliases can be found.
+func (p *Parser) ReplaceAliases(ctx context.Context, aliasBackend aliases.AliasBackend, addrs []string) ([]string, error) {
```
darioAnongba, line 48: the fact that you have to pass the AliasBackend here is a strong argument as to why this doesn't belong in a separate component

darioAnongba (approved):

src/aliases/api/handlers/handlers_test.go
darioAnongba: Good one, I didn't know about this standard

src/aliases/parser/parser.go
dolanor: yep. Fixed.

PR #297, opened 2021-09-09
fix message signature + verify

6l20:
## PR description

- Change message format to string, now accepts hexaString w/ 0x and wo/ 0x
- Fixed EIP 191 format (missing ":" char)
- Fixed recID value with x+27
- Changed recover recID with x-27

## Changelog

### DIFF
--- a/src/stores/api/types/eth1.go
+++ b/src/stores/api/types/eth1.go
@@ -24,7 +24,7 @@ type UpdateEthAccountRequest struct {
 }
 
 type SignMessageRequest struct {
-	Message hexutil.Bytes `json:"message" validate:"required" example:"0xfeade..." swaggertype:"string"`
+	Message string `json:"message" validate:"required" example:"0x536f6d652064617461 or 536f6d652064617461 (hexaString)" swaggertype:"string"`
 }
 
 type SignTypedDataRequest struct {
--- a/src/stores/connectors/ethereum/sign.go
+++ b/src/stores/connectors/ethereum/sign.go
@@ -40,7 +40,10 @@ func (c Connector) Sign(ctx context.Context, addr common.Address, data []byte) (
 func (c Connector) SignMessage(ctx context.Context, addr common.Address, data []byte) ([]byte, error) {
 	logger := c.logger.With("address", addr)
 
-	signature, err := c.sign(ctx, addr, crypto.Keccak256([]byte(getEIP191EncodedData(data))))
+	toHash := []byte(getEIP191EncodedData(data))
+	hash := crypto.Keccak256(toHash)
+
+	signature, err := c.sign(ctx, addr, hash)
 	if err != nil {
 		return nil, err
 	}
@@ -227,6 +230,8 @@ func (c Connector) sign(ctx context.Context, addr common.Address, data []byte) (
 		}
 
 		if bytes.Equal(crypto.FromECDSAPub(recoveredPubKey), acc.PublicKey) {
+			// add + 27 to recID
+			appendedSignature[len(appendedSignature)-1] += 27
 			return appendedSignature, nil
 		}
 	}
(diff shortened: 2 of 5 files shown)


src/stores/connectors/ethereum/utils.go
```diff
@@ -27,7 +25,7 @@ func getEIP712EncodedData(typedData *core.TypedData) (string, error) {
 }
 
 func getEIP191EncodedData(msg []byte) string {
```
ggarri, line 27: Minor: This method should return the []byte directly as returned values is always used as []byte not as string

src/stores/connectors/ethereum/sign.go
```diff
@@ -227,6 +230,8 @@ func (c Connector) sign(ctx context.Context, addr common.Address, data []byte) (
 		}
 
 		if bytes.Equal(crypto.FromECDSAPub(recoveredPubKey), acc.PublicKey) {
+			// add + 27 to recID
+			appendedSignature[len(appendedSignature)-1] += 27
```
ggarri, line 234: what is the reasoning behind V changed value +27? V is calculated based on the ChainID and it should be the Signer struct  who calculates that based on the chainID you can see next:
```
func (s EIP155Signer) SignatureValues(tx *Transaction, sig []byte) (R, S, V *big.Int, err error) {
   R, S, V, err = HomesteadSigner{}.SignatureValues(tx, sig)
   if err != nil {
      return nil, nil, nil, err
   }
   if s.chainId.Sign() != 0 {
      V = big.NewInt(int64(sig[64] + 35))
      V.Add(V, s.chainIdMul)
   }
   return R, S, V, nil
}
```

src/stores/api/types/eth1.go
```diff
@@ -24,7 +24,7 @@ type UpdateEthAccountRequest struct {
 }
 
 type SignMessageRequest struct {
-	Message hexutil.Bytes `json:"message" validate:"required" example:"0xfeade..." swaggertype:"string"`
+	Message string `json:"message" validate:"required" example:"0x536f6d652064617461 or 536f6d652064617461 (hexaString)" swaggertype:"string"`
```
ggarri, line 27: why did you need to change the input type to string if you still expecting a hex value?

src/stores/connectors/ethereum/utils.go
```diff
@@ -27,7 +25,7 @@ func getEIP712EncodedData(typedData *core.TypedData) (string, error) {
...
-	return fmt.Sprintf("\x19Ethereum Signed Message\n%d%v", len(msg), hexutil.Encode(msg))
+	return fmt.Sprintf("\x19Ethereum Signed Message:\n%d%v", len(msg), string(msg))
```
ggarri, line 28: Did you find out where is documented which format message should be appended to the message?

src/stores/connectors/ethereum/sign.go
ggarri: okay, I found the reasoning behind the +27 in this EIP-155 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md, apparently in case of signing for not specific chain and to prevent replay attacks we need to do V {0,1} + 27.
In any case being that the case I would rather create a specific method to make it more explicit
func eip151(signature []byte) []byte {
	l := len(signature)
	hl := l / 2

	R := new(big.Int).SetBytes(signature[:hl])
	S := new(big.Int).SetBytes(signature[hl:l-1])
    V := new(big.Int).SetBytes(signature[l-1])
    V := V + 27
	return append(append(R.Bytes(), S.Bytes()...), V.Bytes()...)
}

src/stores/api/types/eth1.go
6l20: Because in the end I had the intention to propose accepting human readable messages, but this is out of scope. Messages such as 'Hello world' for instance. That should be supported too.

src/stores/connectors/ethereum/utils.go
6l20: No, I took this implementation out of the ethers.js lib

PR #305, opened 2021-09-09
Fix sign message endpoints (sign message + sign typed data)

darioAnongba:
Fix sign message endpoints (sign message + sign typed data):
- Remove Verify from endpoint since that function does not make sense anymore (same as removal of Sign)
- Modify SignMessage and SignTypedData to sign with HomesteadSigner
- Modify VerifyMessage and VerifyTypedData functions to verify Homestead signatures


src/stores/connectors/ethereum/recover.go
```diff
@@ -11,9 +11,8 @@ import (
 func (c Connector) ECRecover(_ context.Context, data, sig []byte) (ethcommon.Address, error) {
 	pubKey, err := crypto.SigToPub(crypto.Keccak256(data), sig)
 	if err != nil {
-		errMessage := "failed to recover public key, please verify your signature and payload"
-		c.logger.WithError(err).Error(errMessage)
-		return ethcommon.Address{}, errors.InvalidParameterError(errMessage)
+		c.logger.WithError(err).Error("failed to recover public key, please verify your signature and payload")
+		return ethcommon.Address{}, errors.InvalidParameterError(err.Error())
```
darioAnongba, line 15: here we want to get the error from go-ethereum since it's a 4xx error that we can return to the user.

src/stores/connectors/ethereum/sign.go
```diff
@@ -248,6 +248,22 @@ func eeaHash(object interface{}) (hash common.Hash, err error) {
...
+func (c Connector) signHomestead(ctx context.Context, addr common.Address, data []byte) ([]byte, error) {
+	signature, err := c.sign(ctx, addr, data)
+	if err != nil {
+		return nil, err
+	}
+
+	r, s, v, err := types.HomesteadSigner{}.SignatureValues(nil, signature)
+	if err != nil {
+		errMessage := "failed to recover homestead signature values"
+		c.logger.Error(errMessage)
```
ggarri, line 260: Missing `WithError`

ggarri (approved):

PR #307, opened 2021-09-10
Alias e2e tests

dolanor:
Add e2e tests for Alias.

The biggest change on the service side was to create an empty slice for the ListAliases so the JSON returns `[]` and not `null`.


tests/e2e/alias_test.go
```diff
@@ -0,0 +1,242 @@
...
+		reg:    types.RegistryName("JPM-" + randID),
+		key:    types.AliasKey("GoldmanSachs-" + randID),
+		val:    `["ROAZBWtSacxXQrOe3FGAqJDyJjFePR5ce4TSIzmJ0Bc=","2T7xkjblN568N1QmPeElTjoeoNT4tkWYOJYxSMDO5i0="]`,
+		newVal: `["ZOAZBWtSacxXQrOe3FGAqJDyJjFePR5ce4TSIzmJ0Bc=","2T7xkjblN568N1QmPeElTjoeoNT4tkWYOJYxSMDO5i0="]`,
+	}
+}
+
+func (s *aliasTestSuite) TestFull() {
+	fakeAlias := s.fakeAlias()
+	s.Run("create", func() {
```
ggarri, line 70: Please use the e2e standards we use in other e2e test units https://github.com/ConsenSys/quorum-key-manager/blob/main/tests/e2e/ethereum_test.go#L112
```
- should {action} successfully
- should failed [with error type] {when|if} {failure scenario} 
```

ggarri (approved):
Please make sure you update the e2e test unit message to align with the rest of test units

dolanor:
Updated test message.
Waiting for end of CI before merging.

PR #308, opened 2021-09-10
Implement EIP-1559 HTTP

darioAnongba:
Implement EIP-1559 HTTP


src/stores/api/formatters/ethereum.go
```diff
@@ -60,11 +62,55 @@ func FormatSignTypedDataRequest(request *types.SignTypedDataRequest) *signer.Typ
...
+			ChainID:    tx.ChainID.ToInt(),
+			Nonce:      uint64(tx.Nonce),
+			GasPrice:   tx.GasPrice.ToInt(),
+			Gas:        uint64(tx.GasLimit),
+			To:         tx.To,
+			Value:      tx.Value.ToInt(),
+			Data:       tx.Data,
+			AccessList: tx.AccessList,
+		}
+	case "", types.DynamicFeeTxType:
```
ggarri, line 89: MINOR: I would rather not allow empty values to reach this switch and set a default value on the API service method

src/stores/api/types/ethereum.go
```diff
@@ -3,11 +3,20 @@ package types
...
+	"github.com/ethereum/go-ethereum/core/types"
+
 	"github.com/ethereum/go-ethereum/common"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+// Transaction types
+const (
+	LegacyTxType     = "LEGACY"
```
ggarri, line 15: minor: Did you copy those string values from other APIs as standard? If not, I would be more incline for using lower case values as it appears to be more conventional in other enum values we have, for instance PrivateTxType

src/stores/api/formatters/ethereum.go
```diff
@@ -60,11 +62,55 @@ func FormatSignTypedDataRequest(request *types.SignTypedDataRequest) *signer.Typ
 	return typedData
 }
 
-func FormatTransaction(tx *types.SignETHTransactionRequest) *ethtypes.Transaction {
-	if tx.To == nil {
-		return ethtypes.NewContractCreation(uint64(tx.Nonce), tx.Value.ToInt(), uint64(tx.GasLimit), tx.GasPrice.ToInt(), tx.Data)
+func FormatTransaction(tx *types.SignETHTransactionRequest) (*ethtypes.Transaction, error) {
```
ggarri, line 65: In my opinion we should fail in the following cases:
- user set `GasPrice` and it does not use type `DynamicFeeTxType`
- user set `AccessList` and it does use Legacy
- ....

deps/geth/config/keystore/.gitkeep
```diff
@@ -0,0 +1 @@
+These addresses are the same as the default accounts available in Orchestrate
```
dolanor, line 1: Is this some documentation about this directory?
Maybe we should rename as README.md so it can be more useful (and will still avoid git from removing the dir)

src/stores/api/types/ethereum.go
darioAnongba: replaced with lower case values

src/stores/api/formatters/ethereum.go
darioAnongba: it's better to use the empty string in the switch IMO since the input param of this function is the request type, in which the value can be empty. This centralizes the decision that the default value is Dynamic fee also in the tests.

src/stores/api/formatters/ethereum.go
darioAnongba: the behavior of nodes (the standard) is to ignore the fields that do not make sense given the type of tx. If you provide gasPrice but the tx is EIP-1559 (dynamic fee) then it ignores gasPrice. Same for accessList. Personally I prefer that behavior

src/stores/api/types/testutils/faker.go
```diff
@@ -94,18 +96,61 @@ func FakeSignTypedDataRequest() *types.SignTypedDataRequest {
...
+			ChainID:         hexutil.Big(*hexutil.MustDecodeBig("0x1")),
+			Data:            hexutil.MustDecode("0xfeee"),
+			AccessList: []ethtypes.AccessTuple{
+				{
+					Address:     toAddress,
+					StorageKeys: []common.Hash{common.HexToHash("0xfeee")},
+				},
+			},
+		}
 	}
```
dolanor, line 151: minor: It's hard to see in this case which is the difference between the returns.
Can we just have what's different in the cases, and after the switch the return with every parameters set correctly.
And a default case to `return nil` if no match is found?

pkg/client/ethereum.go
```diff
@@ -185,17 +185,6 @@ func (c *HTTPClient) ECRecover(ctx context.Context, storeName string, req *types
 	return parseStringResponse(response)
 }
 
-func (c *HTTPClient) Verify(ctx context.Context, storeName string, req *types.VerifyRequest) error {
```
dolanor, line 188: What is the reason to delete this?

deps/geth/config/keystore/.gitkeep
darioAnongba: renamed to readme

pkg/client/ethereum.go
darioAnongba: Verify endpoint was removed but we forgot to remove it from the client.

src/stores/api/types/testutils/faker.go
darioAnongba: true! I mixed this faker with the formatter and didn't realise the type is the same in all 3 switches. Done.

src/stores/api/formatters/ethereum.go
darioAnongba: discussed

ggarri (approved):

PR #314, opened 2021-09-14
Implement EIP-1559 JSON-RPC

darioAnongba:
Implement EIP-1559 JSON-RPC


pkg/ethereum/call_msg.go
```diff
@@ -4,18 +4,23 @@ import (
...
-	From     *ethcommon.Address // the sender of the 'transaction'
-	To       *ethcommon.Address // the destination contract (nil for contract creation)
-	Gas      *uint64            // if 0, the call executes with near-infinite gas
-	GasPrice *big.Int           // wei <-> gas exchange ratio
-	Value    *big.Int           // amount of wei sent along with the call
-	Data     *[]byte            // input data, usually an ABI-encoded contract method invocation
+	From       *ethcommon.Address // the sender of the 'transaction'
+	To         *ethcommon.Address // the destination contract (nil for contract creation)
+	Gas        *uint64            // if 0, the call executes with near-infinite gas
+	GasPrice   *big.Int           // wei <-> gas exchange ratio
```
ggarri, line 18: do you know why are we using big.Int here and hexutil.Big in the rest of Msg structs?

src/nodes/interceptor/eth_send_transaction.go
```diff
@@ -2,19 +2,29 @@ package interceptor
...
 	proxynode "github.com/consensys/quorum-key-manager/src/nodes/node/proxy"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 )
 
 func (i *Interceptor) ethSendTransaction(ctx context.Context, msg *ethereum.SendTxMsg) (*ethcommon.Hash, error) {
-	i.logger.Debug("sending ETH transaction")
+	switch {
+	case msg.IsPrivate():
+		return i.sendPrivateTx(ctx, msg)
+	case msg.GasPrice != nil:
```
ggarri, line 18: I think, even if complexity is low, having a helper function as "isLegacy" might be useful to make our code more verbose and better encapsulate

src/nodes/interceptor/eth_sign_transaction.go
```diff
@@ -57,7 +53,14 @@ func (i *Interceptor) ethSignTransaction(ctx context.Context, msg *ethereum.Send
 	if msg.IsPrivate() {
 		sig, err = store.SignPrivate(ctx, msg.From, msg.TxDataQuorum())
 	} else {
-		sig, err = store.SignTransaction(ctx, msg.From, chainID, msg.TxData())
+		var txType int
+		if msg.GasPrice != nil {
```
ggarri, line 57: Replace by the pending helper function `msg.isLegacy()` so we have same conditions applied on everywhere

src/nodes/interceptor/eth_sign_transaction.go
```diff
@@ -21,12 +23,6 @@ func (i *Interceptor) ethSignTransaction(ctx context.Context, msg *ethereum.Send
 		return nil, jsonrpc.InvalidParamsError(errors.InvalidParameterError(errMessage))
 	}
 
-	if msg.GasPrice == nil {
```
ggarri, line 24: Should we check here the following
```
if msg.isLegacy() {
   if msg.GasPrice == nil {
       // error
   }
} else {
   if msg.GasFeeCap == nil {
       // error
   }
}
```

src/nodes/interceptor/eth_sign_transaction.go
```diff
@@ -57,7 +53,14 @@ func (i *Interceptor) ethSignTransaction(ctx context.Context, msg *ethereum.Send
 	if msg.IsPrivate() {
```
ggarri, line 53: I would rather replace this code by a switch stament to make it more readable:
```
switch
  case msg.IsPrivate():
  case msg.IsLegacy():
  default:
    // DynamicFeeTxType

File: src/nodes/interceptor/eth_send_transaction.go [hunk 6]
```diff
@@ -37,65 +47,183 @@ func (i *Interceptor) ethSendTransaction(ctx context.Context, msg *ethereum.Send
...
+	i.logger.Info("quorum private transaction sent successfully", "tx_hash", hash)
+	return &hash, nil
+}
+
+func (i *Interceptor) sendLegacyTx(ctx context.Context, msg *ethereum.SendTxMsg) (*ethcommon.Hash, error) {
+	i.logger.Debug("sending ETH legacy transaction")
+
+	sess := proxynode.SessionFromContext(ctx)
+
+	if msg.GasPrice == nil {
```
ggarri, line 100: I like the encapsulation you did with `fillNonce()`, could you do the same here?

src/nodes/interceptor/eth_send_transaction.go
```diff
@@ -37,65 +47,183 @@ func (i *Interceptor) ethSendTransaction(ctx context.Context, msg *ethereum.Send
...
 
-	if msg.IsPrivate() {
-		if msg.Data == nil {
-			msg.Data = new([]byte)
-		}
-
-		var privateFrom string
-		if msg.PrivateFrom != nil {
-			privateFrom = *msg.PrivateFrom
+	if msg.Gas == nil {
```
ggarri, line 110: I like the encapsulation you did with `fillNonce()`, could you do the same here?

src/nodes/interceptor/eth_send_transaction.go
darioAnongba: that is not true. The behavior of the standard is: "If the gasPrice is specified, the transaction is a legacy transaction". Whether you specify other params or not is irrelevant.

src/nodes/interceptor/eth_sign_transaction.go
darioAnongba: these checks are not correct

src/nodes/interceptor/eth_sign_transaction.go
darioAnongba: done

src/nodes/interceptor/eth_sign_transaction.go
darioAnongba: done

src/nodes/interceptor/eth_send_transaction.go
darioAnongba: done

src/nodes/interceptor/eth_send_transaction.go
darioAnongba: done

pkg/ethereum/call_msg.go
darioAnongba: The architecture of the JSON-RPC is not correct as everything is mixed between infra, service and business layer inside `pkg/ethereum`. Some calls are done to the underlying node (infra). Some are requests (JSON received in HTTP) and some are internal types (entities). We should fix this mess at some point. Here I only followed what was already implemented.

ggarri: If that is the case, would you mind to add a TODO with your findings here? Once you do that you can resolve this

src/nodes/interceptor/eth_sign_transaction.go
darioAnongba: discussed as gasPrice = 0 is valid and maxFee = 0 as well.

src/nodes/interceptor/eth_send_transaction.go
darioAnongba: discussed and comment added to explain that pre-london and post-london check

ggarri (approved):

PR #316, opened 2021-09-15
feat(docker): add extra docker registry & cleanup docker setup

julien-marchand:
## Changelog


Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG VERSION=nonroot
 
 # Build the manager binary
-FROM golang:1.15 as builder
+FROM golang:1.17 as builder
```
ggarri, line 4: why are you bumping this?

docker-compose.dev.yml
```diff
@@ -0,0 +1,106 @@
+version: "3.7"
```
ggarri, line 1: Please make sure you do git mv, instead of creating a new file, otherwise we lost historial changes

Dockerfile
julien-marchand: No technical reason, just to follow the latest Go version.

docker-compose.dev.yml
julien-marchand: That's what I did and what my local git is saying. Probably an issue because the old name of one file is the same as the new name of the other

Dockerfile
ggarri: I would rather postpone this till we get more time to properly test it on that version, JIC

docker-compose.dev.yml
ggarri: I think you will need to revert this and then do:
git mv docker-compose.yml docker-compose.dev.yml
git mv docker-compose.latest.yml docker-compose.yml

ggarri (approved):

PR #323, opened 2021-09-16
rename hashicorp plugin to use new repo

ggarri:
## Changelog


deps/hashicorp/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:3.14
```
darioAnongba, line 1: what about just using `3` and not worrying about the `minor` ?

darioAnongba (approved):

6l20 (approved):

deps/hashicorp/Dockerfile
ggarri: I tend to not rely on automatic changes even if they are minor since they can happen at any time with no plannification and breaks unexpectcly

PR #341, opened 2021-09-24
Hashicorp Vault TLS server and client

darioAnongba:
Hashicorp Vault TLS server and client


deps/hashicorp/docker-compose.yml
```diff
@@ -23,7 +24,10 @@ services:
     <<: *container-common
     build: .
     environment:
-      VAULT_ADDR: ${VAULT_ADDR-http://hashicorp:8200}
+      VAULT_ADDR: ${VAULT_ADDR-https://hashicorp:8200}
```
ggarri, line 27: I would rather not make TLS the default dev setup since it introduces another area of potential additional on our daily work instead I would rather have `vault-tls` service and everything in place in decide whether or not to use TLS only when we intend to test QKM with it

src/stores/api/types/stores.go
```diff
@@ -0,0 +1,22 @@
+package types
+
+import "time"
+
+// TODO: Keep this type as an example of the request type when a manifest is read
+type CreateHashicorpStoreRequest struct {
```
ggarri, line 6: Since this is not use and not decided yet it would be used at some point, I would propose to remove it

src/stores/entities/specs.go
```diff
@@ -0,0 +1,23 @@
+package entities
+
+import "time"
+
+// HashicorpSpecs is the specs format for a Hashicorp Vault client
+// TODO: Remove annotations when type from service layer is used
+type HashicorpSpecs struct {
```
ggarri, line 7: Since you extracted this Specs, which I believe it is a good decision, I would expect to see the same for AKV and AWS, or at least have a ticket to do so in the backlog

src/stores/api/types/stores.go
darioAnongba: ah yes forgot to remove

src/stores/entities/specs.go
darioAnongba: https://app.zenhub.com/workspaces/orchestrate-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/169

ggarri (approved):

src/stores/manager/keys/hashicorp.go
```diff
@@ -49,21 +38,6 @@ func NewHashicorpKeyStore(specs *HashicorpKeySpecs, logger log.Logger) (*hashico
 				logger.Warn("token watcher has exited gracefully")
 			}
 		}()
-
-		// We wait for the token to be set before we continue
-		for currRetries := 1; currRetries <= maxRetries; currRetries++ {
```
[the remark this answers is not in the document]
darioAnongba: discussed

PR #375, opened 2021-10-07
Refactor Manifest reader

darioAnongba:
Make manifest reader synchronous and not a service anymore but a client in `infra`


src/app.go
```diff
@@ -7,48 +7,42 @@ import (
...
+	Manifest *manifests.Config
 }
 
 func New(cfg *Config, logger log.Logger) (*app.App, error) {
 	// Create app
 	a := app.New(&app.Config{HTTP: cfg.HTTP}, logger.WithComponent("app"))
 
 	// Register Service Configuration
-	err := a.RegisterServiceConfig(cfg.Manifests)
+	err := a.RegisterServiceConfig(cfg.Auth)
```
ggarri, line 30: IRCC auth config should also include `cfg.manifest` as Roles are loaded from manifest files

src/nodes/manager/service.go
```diff
@@ -169,22 +154,21 @@ func (m *BaseManager) List(_ context.Context, userInfo *authtypes.UserInfo) ([]s
...
 
-func (m *BaseManager) load(ctx context.Context, mnf *manifest.Manifest) error {
+func (m *BaseManager) createNodes(ctx context.Context, mnf *manifest.Manifest) error {
 	m.mux.Lock()
 	defer m.mux.Unlock()
 
-	logger := m.logger.With("kind", mnf.Kind, "name", mnf.Name)
+	logger := m.logger.With("name", mnf.Name)
 
 	if _, ok := m.nodes[mnf.Name]; ok {
```
ggarri, line 163: I cannot understand the propose of this check here, did you?

ggarri (approved):

Issue #373, opened 2021-10-06
[Hashicorp client] TLS config error must not be ignored


6l20:
### Describe the bug

Errors occurring during TLS configuration of Hashicorp client are ignored.

### Configuration

Pass a non existing file as the client certificate.

### Steps to reproduce

In manifest config enter wrong values for clientCert or no value when clientKey is a correct value

**Actual result**

No error

**Expected result**

An error that says file is not found or missing

### Additional context

**Other**

PR #376, opened 2021-10-08
Fix/tls config hashicorp #373

6l20:
## PR description
- No longer ignore Hashicorp instanciation / config errors
- Add warning log when VAULT_SKIP_VERIFY set
## Fixed Issue(s)
#373

## Changelog


src/stores/manager/keys/hashicorp.go
```diff
@@ -22,6 +22,10 @@ func NewHashicorpKeyStore(specs *entities.HashicorpSpecs, logger log.Logger) (*h
 		return nil, errors.ConfigError(errMessage)
 	}
 
+	if cfg.SkipVerify {
```
darioAnongba, line 25: @sylvain.decourval you need to add this code to the function Create in src/stores/connectors/stores/create.go instead of where it's now. Should be added in both switches of HashicorpKeys and HashicorpSecrets.
The check will be by checking the specs spec.SkipVerify

PR #387, opened 2021-10-14
Best effort import script

darioAnongba:
Best effort import script


cmd/imports/ethereum.go
```diff
@@ -36,24 +36,25 @@ func ImportEthereum(ctx context.Context, db database.ETHAccounts, mnf *manifest.
 	for _, id := range storeIDs {
 		key, err := store.Get(ctx, id)
 		if err != nil {
-			return err
+			continue
```
ggarri, line 39: IMO we should warning to notify users something unexpected is happening

cmd/imports/secrets.go
```diff
@@ -35,18 +35,18 @@ func ImportSecrets(ctx context.Context, db database.Secrets, mnf *manifest.Manif
 	for _, id := range difference(storeIDs, dbIDs) {
 		secret, err := store.Get(ctx, id, "")
 		if err != nil {
-			return err
+			continue
```
ggarri, line 38: I would rather warning instead of just ignoring

PR #420, opened 2021-11-12
Refactor manifests and separate vaults from stores

darioAnongba:
Refactor manifest types and separate vaults from stores.
- Read manifest file only once and distribute the manifests to the corresponding service layers
- Define clear request types for all manifest kinds: Store, Role, Vault and Node
- Remove all managers and use dependency injection instead by creating infra resources that are injected in the domains

Missing:
- changelog
- verify that e2e don't break except for the loading of manifests

This PR follows the design: https://docs.google.com/document/d/17-RSVbYAbufhhaQ3E2V0v5s0kiduiSeLEihF5XrZ4u4/edit?usp=sharing


deps/config/manifests/nodes.yml.template
```diff
@@ -1,17 +1,15 @@
...
-      addr: http://validator1:8545
-- kind: Node
-  name: geth-node
+    name: besu-node
+    rpc: http://validator1:8545
+
+- kind: node
   specs:
-    rpc:
-      addr: http://geth:8545
```
ggarri, line 17: This is a breaking change, we should deliverate this more before. DId you consider other RPC alternatives to be available such IPC?

deps/config/manifests/roles.yml.template
```diff
@@ -1,22 +1,25 @@
...
+- kind: role
   specs:
-    permission:
+    name: guest
+    permissions:
       - "read:*"
-- kind: Role
-  name: signer
+
+- kind: role
```
ggarri, line 13: What is the reasoning behind making Kind lowercase? this is another breaking change to be careful of and do it only in case we have strong reasons for it

src/auth/service/roles/roles.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/consensys/quorum-key-manager/src/auth/entities"
+	"sync"
+
+	"github.com/consensys/quorum-key-manager/src/auth"
+	"github.com/consensys/quorum-key-manager/src/infra/log"
+)
+
+type Interactor struct {
```
ggarri, line 13: Why did you name it `Interactor` ? isn't this just a InMemoryRoles struct?

src/nodes/service/nodes/nodes.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/consensys/quorum-key-manager/src/nodes/entities"
+	proxynode "github.com/consensys/quorum-key-manager/src/nodes/node/proxy"
+	"github.com/consensys/quorum-key-manager/src/stores"
+	"sync"
+
+	"github.com/consensys/quorum-key-manager/src/auth"
+	"github.com/consensys/quorum-key-manager/src/infra/log"
+)
+
+type Interactor struct {
```
ggarri, line 17: WHy interactor?

src/stores/app/service.go
```diff
@@ -1,55 +1,56 @@
...
-	// Create and register the stores service
-	storesConnector := stores.NewConnector(*authManager, db, logger)
-	utilsConnector := utils.NewConnector(logger)
-	api.New(storesConnector, utilsConnector).Register(a.Router())
-
-	service := storesmanager.New(storesConnector, manifestReader, db, logger)
-	err = a.RegisterService(service)
-	if err != nil {
-		return err
+	manifestStoreHandler := manifest.NewStoresHandler(storesConnector) // Manifest reading is synchronous, similar to a config file
```
ggarri, line 47: StoreHandler would inject to inject VaultHandler to grab the vault clients into the corresponding Stores, do I missing something?

src/stores/app/service.go
```diff
@@ -1,55 +1,56 @@
...
-	err = a.Service(authManager)
-	if err != nil {
-		return err
+	// Service layer
+	router := a.Router()
+	http.NewStoresHandler(storesConnector).Register(router)
+	http.NewUtilsHandler(utilsConnector).Register(router)
+
+	manifestVaultHandler := manifest.NewVaultsHandler(vaultsConnector) // Manifest reading is synchronous, similar to a config file
+	for _, mnf := range manifests[entities.VaultKind] {
```
ggarri, line 40: IMO it would be better to separate the Store service registration from its state initialization (manifest loading)

src/auth/api/http_middlewares/accesslog.go
```diff
@@ -1,4 +1,4 @@
-package middlewares
+package http_middlewares
```
ggarri, line 1: MINOR: Why did you decide on renaming this? IMHO we could either keep `middlewares` or rename it `http` but composed names are not very common in Go as far as I expericed

src/auth/app/service.go
```diff
@@ -3,112 +3,72 @@ package app
...
-	if err != nil {
-		return err
-	}
+	rolesService := roles.New(logger)
 
-	compositionMiddleware, err := createMiddlewares(cfg, logger)
-	if err != nil {
-		return err
+	// Service layer
+	manifestRolesHandler := manifest.NewRolesHandler(rolesService) // Manifest reading is synchronous, similar to a config file
```
ggarri, line 39: we are mixing service initialization with service state initialization in all services. I would rather do it in two independent steps. Most likely we should look for a common space for all manifestHandlers where the state of all services are managed, WDYT?

src/app.go
```diff
@@ -20,57 +25,124 @@ type Config struct {
...
-	err = nodesapp.RegisterService(a, logger.WithComponent("nodes"))
+	_, err = nodesapp.RegisterService(ctx, a, logger.WithComponent("nodes"), manifests, authService, storesService, aliasService)
 	if err != nil {
 		return nil, err
 	}
 
 	return a, nil
 }
+
+func getAPIKeys(ctx context.Context, cfg *csv.Config, logger log.Logger) (map[string]*authtypes.UserClaims, error) {
```
ggarri, line 86: Auth service constructors(getAPIKEys, getJWTValidator and getRootCAs) must be located under Auth package for better encapsulation of responsibilities instead of having them in `app.go`.

src/app.go
```diff
@@ -20,57 +25,124 @@ type Config struct {
...
+	if cfg != nil {
```
ggarri, line 87: We discuss this pattern in the past and I thought we agreed on the fact we should reduce the `!= nil` checks on constructors and make caller to decide whether or not an object must be instanciated. Using this impl we will increase the complexity of our UTs

src/auth/app/service.go
```diff
@@ -3,112 +3,72 @@ package app
...
+	manifests map[string][]entities2.Manifest,
+	jwtValidator jwt.Validator,
+	apikeyClaims map[string]*authtypes.UserClaims,
+	rootCAs *x509.CertPool,
+) (*roles.Interactor, error) {
+	// Business layer
+	// TODO: Create authorizator service here
+
+	var authenticatorService *authenticator.Authenticator
+	if jwtValidator != nil || apikeyClaims != nil || rootCAs != nil {
```
ggarri, line 32: Having services which might be nil depending on the settings of the user appears to be a bad practices since it forces us to do checks over our service execution flow. I would rather encapsulate the fact not authentication methods were enabled under the authService itself

src/auth/service.go
```diff
@@ -22,14 +22,10 @@ type Authorizator interface {
...
-type Manager interface {
-	// Role returns role
-	Role(name string) (*entities.Role, error)
-
-	// Roles returns roles
-	Roles() ([]string, error)
-
-	// UserPermissions Extract User Permissions from UserInfo
-	UserPermissions(info *entities.UserInfo) []entities.Permission
+// Roles allows managing policies and roles
```
ggarri, line 25: // Roles allows managing permissions and roles

src/auth/service/roles/roles.go
```diff
@@ -0,0 +1,51 @@
...
+	mux    sync.RWMutex
+	roles  map[string]*entities.Role
+	logger log.Logger
+}
+
+var _ auth.Roles = &Interactor{}
+
+func New(logger log.Logger) *Interactor {
```
ggarri, line 21: I would rather have a single file with all its methods together since there is not so much complexity on the methods itself for separation

src/infra/manifests/yaml/reader.go
```diff
@@ -86,3 +91,15 @@ func (r *Reader) buildMessages(fp string) ([]*manifest.Manifest, error) {
 
 	return mnfs, nil
 }
+
+func fillMap(mnfs []entities.Manifest, manifestsMap map[string][]entities.Manifest) {
```
ggarri, line 95: Minor: Method name is not very descriptive, I would suggest something like:
```
func aggregateManifestKind(mnfs []entities.Manifest) map[string][]entities.Manifest {
....
```

src/stores/api/http/context.go
```diff
@@ -1,4 +1,4 @@
-package handlers
```
ggarri, line 1: If you don't mind you should apply same rename change to `aliases` package for better consistency

(I added the comment here because I didn't find better place to do so)

src/stores/connectors/stores/create_key.go
```diff
@@ -2,177 +2,83 @@ package stores
...
+	var err error
+	switch {
+	case vault != "":
+		store, err = c.createKeyStore(ctx, vault)
+		if err != nil {
+			return err
 		}
+	case secretStore != "":
+		// TODO: Uncomment when authManager no longer a runnable
+		// permissions := c.authManager.UserPermissions(userInfo)
```
ggarri, line 43: What would it be the propose of this action here?

src/stores/connectors/vaults/create_aws.go
```diff
@@ -0,0 +1,25 @@
+package vaults
```
ggarri, line 1: Vaults are not connectors since they are not containing any business logic. They are wrappers around the infra clients so I would rather place them elsewhere, I would suggest `/src/stores/vaults` since it is the same kind of level of components of `/src/stores/store`

deps/config/manifests/roles.yml.template
darioAnongba: now everything is lowercase

src/auth/service/roles/roles.go
darioAnongba: called Roles

src/auth/api/http_middlewares/accesslog.go
darioAnongba: resolved

src/stores/connectors/stores/create_key.go
darioAnongba: Add RBAC for vaults (Resource)

src/stores/app/service.go
darioAnongba: moved

src/auth/app/service.go
darioAnongba: moved

src/app.go
darioAnongba: resolved

src/auth/service.go
darioAnongba: fixed

src/infra/manifests/yaml/reader.go
darioAnongba: done

src/stores/connectors/vaults/create_aws.go
darioAnongba: moved to its own domain

deps/config/manifests/vault.yml.template
```diff
@@ -0,0 +1,49 @@
+- kind: vault
+  type: hashicorp
+  name: hashicorp-kv-v2
+  specs:
+    # allowed_tenants: [tenant1, tenant2]
```
ggarri, line 5: I thought we agreed on moving allowed tenants to first level since it wont be part of the user request payload. Also for simplification as you would have only specs, instead of specs+ config

pkg/json/decode.go
```diff
@@ -20,7 +22,39 @@ func UnmarshalBody(body io.Reader, req interface{}) error {
...
+
+	err = json.Unmarshal(bdata, dest)
+	if err != nil {
+		return err
+	}
+
+	return validateStruct(dest)
+}
+
+func UnmarshalYAML(src, dest interface{}) error {
```
ggarri, line 42: We sre under pkg json, this utility does not seem to belong here

darioAnongba: this is not possible currently because the `json` and `validate` have been merged together inside of the `json` package. I created a ticket to fix this in the futrure and remove the `validate` from the `json`: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/430

src/stores/api/manifest/stores.go
```diff
@@ -0,0 +1,92 @@
+package manifest
+
+import (
+	"context"
+
+	"github.com/consensys/quorum-key-manager/pkg/errors"
+	"github.com/consensys/quorum-key-manager/pkg/json"
+	authtypes "github.com/consensys/quorum-key-manager/src/auth/entities"
+	entities2 "github.com/consensys/quorum-key-manager/src/entities"
```
6l20, line 9: entities2 ?

src/stores/connectors/keys/create.go
```diff
@@ -3,14 +3,16 @@ package keys
 import (
 	"context"
 
+	entities2 "github.com/consensys/quorum-key-manager/src/entities"
```
6l20, line 6: entities2 ?

src/stores/connectors/keys/import.go
```diff
@@ -3,14 +3,16 @@ package keys
 import (
 	"context"
 
+	entities2 "github.com/consensys/quorum-key-manager/src/entities"
```
6l20, line 6: idem

src/stores/connectors/stores/create_key.go
```diff
@@ -2,177 +2,78 @@ package stores
 
 import (
 	"context"
-	"time"
 
+	authtypes "github.com/consensys/quorum-key-manager/src/auth/entities"
+	"github.com/consensys/quorum-key-manager/src/auth/service/authorizator"
+	entities2 "github.com/consensys/quorum-key-manager/src/entities"
```
6l20, line 8: idem

src/init.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+
+	manifests, err := manifestReader.Load(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Note that order is important here as stores depend on the existing vaults, do not use a switch!
```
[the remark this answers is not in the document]
darioAnongba: discussed, this is not related to the manifest but to the fact that stores use vaults so we must create vaults before stores.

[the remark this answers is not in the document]
darioAnongba: discussed, this is not related to the manifest but to the fact that stores use vaults so we must create vaults before stores.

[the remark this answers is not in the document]
darioAnongba: discussed, this is not related to the manifest but to the fact that stores use vaults so we must create vaults before stores.

deps/config/manifests/store.yml.template
```diff
@@ -1,33 +1,55 @@
-- kind: HashicorpSecrets
+- kind: store
```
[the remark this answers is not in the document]
darioAnongba: done

deps/config/manifests/nodes.yml.template
```diff
@@ -1,17 +1,19 @@
-- kind: Node
+- kind: node
```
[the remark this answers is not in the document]
darioAnongba: done (reverted to uppercase for Kind)

pkg/json/validate.go
```diff
@@ -1,7 +1,7 @@
 package json
 
 import (
-	"github.com/consensys/quorum-key-manager/src/stores/entities"
+	entities2 "github.com/consensys/quorum-key-manager/src/entities"
```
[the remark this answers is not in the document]
darioAnongba: entities are now split in the different domains but we have a tech debt to move them to a single folder: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/431

src/stores/api/manifest/stores.go
darioAnongba: entities are now split in the different domains but we have a tech debt to move them to a single folder: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/431

src/stores/connectors/keys/create.go
darioAnongba: entities are now split in the different domains but we have a tech debt to move them to a single folder: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/431

src/stores/connectors/stores/create_key.go
darioAnongba: entities are now split in the different domains but we have a tech debt to move them to a single folder: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/431

src/stores/connectors/keys/import.go
darioAnongba: entities are now split in the different domains but we have a tech debt to move them to a single folder: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/quorum-key-manager/431

src/aliases/interactors/aliases/interactor.go
```diff
@@ -10,29 +10,23 @@ import (
 	"github.com/consensys/quorum-key-manager/src/infra/log"
 )
 
-// We make sure Connector implements aliases.Interactor
-var _ aliases.Interactor = &Interactor{}
-
-// Interactor is the service layer for other service to query.
 type Interactor struct {
```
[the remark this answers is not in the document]
darioAnongba: Indeed `Interactor` is not a good name and won't be kept. It is going to be called simply `Aliases` like in the other domains. I will do that change in my next PR when I work on aliases.

6l20 (approved):

ggarri (approved):

PR #433, opened 2021-12-13
fix panic when not auth are defined

ggarri:
## Changelog

### DIFF
--- a/src/auth/api/http/auth.go
+++ b/src/auth/api/http/auth.go
@@ -3,6 +3,7 @@ package http
 import (
 	"encoding/base64"
 	"net/http"
+	"reflect"
 	"strings"
 
 	"github.com/consensys/quorum-key-manager/pkg/errors"
@@ -29,8 +30,8 @@ func (m *Auth) Middleware(next http.Handler) http.Handler {
 		ctx := r.Context()
 
 		// If no auth method, return wildcard user
-		if m.authenticator == nil {
-			next.ServeHTTP(rw, r.Clone(WithUserInfo(ctx, entities.NewWildcardUser())))
+		if reflect.ValueOf(m.authenticator).IsNil() {
+			next.ServeHTTP(rw, r.WithContext(WithUserInfo(ctx, entities.NewWildcardUser())))
 			return
 		}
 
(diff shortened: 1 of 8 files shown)


src/auth/api/http/auth.go
```diff
@@ -29,8 +30,8 @@ func (m *Auth) Middleware(next http.Handler) http.Handler {
 		ctx := r.Context()
 
 		// If no auth method, return wildcard user
-		if m.authenticator == nil {
-			next.ServeHTTP(rw, r.Clone(WithUserInfo(ctx, entities.NewWildcardUser())))
+		if reflect.ValueOf(m.authenticator).IsNil() {
```
darioAnongba, line 33: Please no usage of reflect in the code, if you need to use reflect then this solution is not the correct one.

src/auth/app/service.go
```diff
@@ -27,7 +27,10 @@ func RegisterService(
 
 	var authenticatorService *authenticator.Authenticator
 	if jwtValidator != nil || apikeyClaims != nil || rootCAs != nil {
+		logger.Info("authentication middleware is enabled")
 		authenticatorService = authenticator.New(jwtValidator, apikeyClaims, rootCAs, logger)
+	} else {
```
darioAnongba, line 32: an `if` statement only to log is an architecture anti-pattern and should normally never happen

src/auth/api/http/auth.go
ggarri: Since we expect an interface here we cannot run the `== nil` here, do you have a different suggestion than reflect?

src/auth/app/service.go
ggarri: I could redefine it as:
```
if authenticatorService == nil 
```

or just not log in this case, but IMO it is important Warn clients auth is not enabled

src/auth/api/http/auth.go
darioAnongba: Like it was implemented before, 2 middlewares, one `Auth` and one `Wildcard` like I had implemented before.

ggarri: Oki, I will do that

PR #437, opened 2021-12-15
Implement alias authorization

darioAnongba:
Implement alias authorization

## Changelog


src/aliases/service/aliases/aliases.go
```diff
@@ -3,21 +3,24 @@ package aliases
...
 	aliasDB    database.Alias
 	registryDB database.Registry
 	logger     log.Logger
+	roles      auth.Roles
 }
 
 var _ aliases.Aliases = &Aliases{}
 
-func New(aliasDB database.Alias, registryDB database.Registry, logger log.Logger) *Aliases {
+func New(aliasDB database.Alias, registryDB database.Registry, rolesService auth.Roles, logger log.Logger) *Aliases {
```
ggarri, line 19: I would expect the Authorizator to be injected here instead of the roles service. I understand the reason, so lets have a chat about it to find out if we could quickly improve this

ggarri (approved):

src/aliases/service/aliases/aliases.go
darioAnongba: yes, I have a good idea on how this needs to be implemented so that the Auth service actually doesn't need to be injected anywhere.

PR #450, opened 2021-12-17
restore encoded version of api keys

ggarri:
## Changelog


src/infra/api-key/csv/reader.go
```diff
@@ -18,15 +15,14 @@ const (
 	csvSeparator         = ','
 	csvCommentsMarker    = '#'
 	csvRowLen            = 4
-	csvHashOffset        = 0
+	csvAPIKeyOffset      = 0
```
6l20, line 18: I dont get why this is called ApiKeyOffset when it is actually the ApiKeyHash that is located here

src/infra/api-key/csv/reader.go
```diff
@@ -65,14 +61,8 @@ func (r *Reader) Load(_ context.Context) (map[string]*entities.UserClaims, error
...
 
-		r.hasher.Reset()
-		_, err = r.hasher.Write([]byte(cells[csvHashOffset]))
-		if err != nil {
-			return nil, fmt.Errorf("failed to hash api key")
-		}
-
-		apiKeyHash := base64.StdEncoding.EncodeToString(r.hasher.Sum(nil))
-		claims[apiKeyHash] = &entities.UserClaims{
+		apiKeySha256 := cells[csvAPIKeyOffset]
```
6l20, line 64: Related to comment above that produces this confusion here

src/auth/service/authenticator/authenticator.go
```diff
@@ -78,10 +77,11 @@ func (authen *Authenticator) AuthenticateAPIKey(_ context.Context, apiKey []byte
 		return nil, fmt.Errorf("failed to hash api key")
 	}
 
-	claims, ok := authen.apiKeyClaims[base64.StdEncoding.EncodeToString(authen.hasher.Sum(nil))]
+	apiKeySha256 := fmt.Sprintf("%x", authen.hasher.Sum(nil))
```
darioAnongba, line 80: why do you need `fmt.Sprintf` here? Seems that you get back []byte and you can simply do `string(authen.hasher.Sum(nil))`

darioAnongba (approved):

src/auth/service/authenticator/authenticator.go
darioAnongba: resolved

src/infra/api-key/csv/reader.go
darioAnongba: done

src/infra/api-key/csv/reader.go
darioAnongba: done

PR #461, opened 2021-12-22
763 sign docker image

ggarri:
## Fixed Issue(s)
#763

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ commands:
     steps:
       - run:
           name: Enable experimental Docker features
-          command: echo 'export DOCKER_CLI_EXPERIMENTAL=enabled' >> $BASH_ENV
+          command: export DOCKER_CLI_EXPERIMENTAL=enabled
       - run:
           name: Docker login
           command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -12,11 +12,23 @@ commands:
...
+            echo "Login in $DOCKER_REGISTRY with user $DOCKER_USER"
             echo $DOCKER_PASS | docker login -u $DOCKER_USER --password-stdin $DOCKER_REGISTRY
-            echo $DOCKER_PASS_2 | docker login -u $DOCKER_USER_2 --password-stdin $DOCKER_REGISTRY_2
+            echo "Login in $DOCKER_REGISTRY_2 with user $DOCKER_USER_RW"
+            echo $DOCKER_PASSWORD_RW | docker login -u $DOCKER_USER_RW --password-stdin $DOCKER_REGISTRY_2
+      - run:
+          name: DCT signing setup
+          command: |
+            mkdir -p $HOME/.docker/trust/private
+            echo $DCT_KEY | base64 --decode > $HOME/.docker/trust/private/$DCT_HASH.key
```
6l20, line 29: Why is the key named DCT_HASH once decoded ?

.circleci/config.yml
```diff
@@ -218,12 +202,57 @@ jobs:
...
+              if [ $TARGETOS != "linux" ]; then
+                echo "Targeted OS not supported $TARGETOS"
+                exit 1
+              fi
+
+              if [ $TARGETARCH == "amd64" ]; then
+                DOCKER_IMAGE_TAG="${DOCKER_IMAGE_REF}"
+              else
+                DOCKER_IMAGE_TAG="${DOCKER_IMAGE_REF}.${TARGETARCH}"
+              fi
```
6l20, line 228: We are loosing the `amd64` tag of the image, maybe we should introduce a new convention here with explicitly tagged arch ?

.circleci/config.yml
ggarri: Those values are being defined in the `dockerhub-opsquorum-dct` CircleCI context by Joshua, I guess, for the opsquorum team. So it is nothing I can do here to rename it

.circleci/config.yml
ggarri: I decided to follow similar strategy than `mysql-server` where the amd64 is consider the default and the arm64 is tagged specifying the arch, although I did a different naming convention which I would be open to discuss if you propose something else

image

https://hub.docker.com/r/mysql/mysql-server/tags?page=1

6l20 (approved):

Issue #462, opened 2022-01-03
Use quorum-hashicorp-plugin image into QKM


ggarri:
The goal of this ticket is to use the `quorum-hashicorp-plugin` image instead of the official vault image.

PR #463, opened 2022-01-03
using vault plugin image

ggarri:
## PR description
Using `quorum-hashicorp-vault-plugin` instead of official image to instantiate the vault server and agent

## Fixed Issue(s)
#462 

## Changelog


tests/acceptance/docker/config/hashicorp/hashicorp.go
```diff
@@ -1,23 +1,17 @@
...
-	Image                 string
-	Port                  string
-	RootToken             string
-	Host                  string
-	PluginSourceDirectory string
+	Image     string
+	Host      string
+	Port      string
+	RootToken string
+	MonthPath string
```
darioAnongba, line 14: `MountPath`

tests/acceptance/docker/config/hashicorp/hashicorp.go
```diff
@@ -1,23 +1,17 @@
...
-	"net/http"
-	"os"
-)
-
-const defaultHashicorpVaultImage = "library/vault:1.8.2"
+const defaultHashicorpVaultImage = "consensys/quorum-hashicorp-vault-plugin:v1.1.4"
 const defaultHostPort = "8200"
 const defaultRootToken = "myRoot"
 const defaultHost = "localhost"
+const defaultMountPath = "orchestrate"
```
darioAnongba, line 7: why `orchestrate` ? Also this mountPath is only for the plugin right? Could be called `pluginMountPath` ?

deps/hashicorp/docker-compose.yml
```diff
@@ -1,126 +1,150 @@
...
     - hashicorp
-
+      
+x-default-variables: &default-variables
+  ROOT_TOKEN_PATH: /vault/token/.root
+  PLUGIN_MOUNT_PATH: quorum
+  KVV2_MOUNT_PATH: secret
+  VAULT_IS_READY: /vault/token/.ready
+  
+x-tls-variables: &tls-variables
```
darioAnongba, line 15: I'm not a fan of these defaults defined in docker-compose because most of the time the different services do not share the same ENV vars so we end up defining ENV vars for services that do not use them. I prefer defining the used ENV vars specifically per service even if there is a bit of copy-paste. 
Currently the agent does not use `PLUGIN_MOUNT_PATH` for example, yet it is defined in the service.

tests/acceptance/docker/config/hashicorp/hashicorp.go
ggarri: Good catch! Updated

tests/acceptance/docker/config/hashicorp/hashicorp.go
ggarri: Updated

deps/hashicorp/docker-compose.yml
ggarri: Actually `PLUGIN_MOUNT_PATH` is used in the agent as well do define the policy capabilities. 

IMO this level of encapsulation fits quiet well into the needs we have here, because:
1. `default-vars` are used by every container
2. `tls-vars` by both tls container (agent and vault)
3. `agent-vars`by both agent containers

darioAnongba: if every var is indeed used by the services where they are defined then it's fine.

darioAnongba (approved):

Issue #432, opened 2021-12-13
Don't print cmd help on manifest errors


ggarri:
### Describe the bug

QKM client prints command help output when an error is encountered in the content of the manifest file

### Steps to reproduce

1. Create an invalid manifest file
2. Start QKM

**Actual result**

CMD help is printed along with the manifest error

**Expected result**

Print manifest error message without any cmd help info

**Logs**

```
key-manager_1  | time="2021-12-13T05:42:40Z" level=fatal msg="execution failed" error="Key: 'Manifest.Kind' Error:Field validation for 'Kind' failed on the 'isManifestKind' tag"
key-manager_1  | 2021-12-13T05:43:41.437Z	ERROR	could not create app	{"error": "Key: 'Manifest.Kind' Error:Field validation for 'Kind' failed on the 'isManifestKind' tag"}
key-manager_1  | Error: Key: 'Manifest.Kind' Error:Field validation for 'Kind' failed on the 'isManifestKind' tag
key-manager_1  | Usage:
key-manager_1  |   key-manager run [flags]
key-manager_1  | 
key-manager_1  | Flags:
key-manager_1  |       --auth-api-key-file string      API key CSV file location.
key-manager_1  |                                       Environment variable: "AUTH_API_KEY_FILE"
key-manager_1  |       --auth-oidc-audience strings    Expected audience ("aud" field) of JWT tokens.
key-manager_1  |                                       Environment variable: "AUTH_OIDC_AUDIENCE"
key-manager_1  |       --auth-oidc-issuer-url string   OpenID Connect issuer server domain (ie. https://quorum-key-manager.eu.auth0.com).
key-manager_1  |                                       Environment variable: "AUTH_OIDC_ISSUER_URL"
key-manager_1  |       --auth-tls-ca string            TLS Authenticator Cert filepath.
key-manager_1  |                                       Environment variable: "AUTH_TLS_CA"
key-manager_1  |       --db-database string            Target Database name
key-manager_1  |                                       Environment variable: "DB_DATABASE" (default "postgres")
key-manager_1  |       --db-host string                Database host
key-manager_1  |                                       Environment variable: "DB_HOST" (default "127.0.0.1")
key-manager_1  |       --db-keepalive duration         Controls the number of seconds after which a TCP keepalive message should be sent 
key-manager_1  |                                       Environment variable: "DB_KEEPALIVE" (default 1m0s)
key-manager_1  |       --db-password string            Database User password
```

**Screenshots**  

[...]

PR #486, opened 2022-01-17
silence usage message if err

ggarri:
## Fixed Issue(s)
#432 

## Changelog

### DIFF
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"fmt"
 	"net/url"
-	"os"
 
 	"github.com/consensys/quorum-key-manager/src/infra/log"
 
@@ -23,10 +22,6 @@ func newMigrateCommand() *cobra.Command {
 		PreRun: func(cmd *cobra.Command, args []string) {
 			preRunBindFlags(viper.GetViper(), cmd.Flags(), "key-manager")
 		},
-		PostRun: func(cmd *cobra.Command, args []string) {
-			// TODO: Identify which error code to return
-			os.Exit(0)
-		},
 	}
 
 	// Register Up command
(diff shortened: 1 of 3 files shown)


cmd/migrate.go
```diff
@@ -23,10 +22,6 @@ func newMigrateCommand() *cobra.Command {
 		PreRun: func(cmd *cobra.Command, args []string) {
 			preRunBindFlags(viper.GetViper(), cmd.Flags(), "key-manager")
 		},
-		PostRun: func(cmd *cobra.Command, args []string) {
```
ggarri, line 26: PostRun is never executed if there is not Run command

cmd/run.go
```diff
@@ -15,14 +15,16 @@ func newRunCommand() *cobra.Command {
...
+			err := runCmd(cmd, args)
+			if err != nil {
+				cmd.SilenceUsage = true
+			}
+			return err
+		},
 		PreRun: func(cmd *cobra.Command, args []string) {
 			preRunBindFlags(viper.GetViper(), cmd.Flags(), "key-manager")
 		},
-		PostRun: func(cmd *cobra.Command, args []string) {
```
ggarri, line 22: PostRun does not trigger when you use RunE, only with Run

cmd/run.go
```diff
@@ -15,14 +15,16 @@ func newRunCommand() *cobra.Command {
 	runCmd := &cobra.Command{
 		Use:   "run",
 		Short: "Run application",
-		RunE:  runCmd,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := runCmd(cmd, args)
+			if err != nil {
+				cmd.SilenceUsage = true
```
darioAnongba, line 21: missing return err?

cmd/run.go
```diff
@@ -15,14 +15,16 @@ func newRunCommand() *cobra.Command {
...
+			}
+			return err
```
darioAnongba, line 23: return nil?

darioAnongba (approved):

cmd/run.go
ggarri: updated

cmd/run.go
ggarri: updated

PR #487, opened 2022-01-17
Rearrange flow

6l20:
## PR description

Rearrange workflow, in order to have e2e running against the remote environment
Make TEST_DATA a parameter

## Changelog


.circleci/config.yml
```diff
@@ -221,28 +233,54 @@ jobs:
...
+      - aws-eks/update-kubeconfig-with-authenticator:
+          cluster-name: << parameters.cluster-name >>
+      - when:
+          condition:
+            equal: [ "true", "${<< parameters.port-frowarding >>}" ]
+          steps:
+            - run:
+                name: Port forward
+                command: |
+                  kubectl port-forward -n << parameters.namespace >> service/quorum-key-manager-quorumkeymanager 8080:8080 8081:8081&
```
ggarri, line 275: Please use background task instead of this
```
      - run:
          name: Background the app
          background: true
          command: |
```

.circleci/config.yml
```diff
@@ -221,28 +233,54 @@ jobs:
...
           command: |
             set +e
-            echo "$MANIFEST_B64" | base64 --decode > ./deps/config/manifests/manifest.yml
-            make deps-tls
-            echo "Wait 10 seconds for dependencies to boot up correctly..."
-            sleep 10
-            make qkm
-            echo "Wait 5 seconds for qkm to boot up correctly..."
-            sleep 5
+            echo "Wait 20 seconds for services to boot up correctly..."
```
ggarri, line 282: Since deployment run async and externally you could quickly do a curl check of your live endpoint here
```
bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' qa.qkm....:8082/live)" != "200" ]]; do sleep 5; done'
```

.circleci/config.yml
```diff
@@ -330,18 +368,54 @@ jobs:
...
+      - prepare_golang
+      - kubernetes/install
+      - aws-eks/update-kubeconfig-with-authenticator:
+          cluster-name: << parameters.cluster-name >>
+      - run:
+          name: Renew Qkm TLS Secret
+          command: |
+            GO111MODULE=off go get github.com/cloudflare/cfssl/cmd/cfssl && go get github.com/cloudflare/cfssl/cmd/cfssljson
+            make generate-pki
+            CA_DIR=tests/e2e/ca
```
ggarri, line 408: IMO, and align to what @darioAnongba raise the other day, we should not generate or depend on pipeline values to deploy our QA/Prod env, they already have a pre define setup and we only trigger the upgrade using the latest image from here, otherwise we are making our CI depending on Helms constrains which might change over time

.circleci/config.yml
```diff
@@ -330,18 +368,54 @@ jobs:
...
+      - go/install:
+          version: 1.16.8
       - checkout
       - set_remote_environment_variables
+      - prepare_golang
+      - kubernetes/install
+      - aws-eks/update-kubeconfig-with-authenticator:
+          cluster-name: << parameters.cluster-name >>
+      - run:
+          name: Renew Qkm TLS Secret
```
ggarri, line 404: IMO, and align to what @darioAnongba raise the other day, we should not generate or depend on pipeline values to deploy our QA/Prod env, they already have a pre define setup and we only trigger the upgrade using the latest image from here, otherwise we are making our CI depending on Helms constrains which might change over time

.circleci/config.yml
```diff
@@ -467,7 +528,39 @@ workflows:
...
+          requires:
+            - deploy-qa
+            # Will build and push a dev image for each main commit
+            # Resulting image is tagged with develop
+            # Images are compliant with linux/amd64 architecture
+      - build-docker-image: # Push Dev image
+          name: push-qkm-dev-image-develop
+          matrix:
+            parameters:
+              docker-image-tag: [ "quorum-key-manager:develop" ]
```
ggarri, line 550: since there is a single argument, there is not need for using matrix here you can pass it ` docker-image-tag` as a regular param

.circleci/config.yml
6l20: Ok, I didn't know this one

.circleci/config.yml
6l20: I am not sure I get what you expect here, these are dependencies needed to access the cluster, that specific part is here to align the qkm existing secret (TLS stuff) with the deployed values. There is no other way I can do that.

.circleci/config.yml
6l20: Well it used to be a matrix, so I kept it.
Will do the change

.circleci/config.yml
ggarri: isn't it possible to mainta same TLS configuration over different deployment?

ggarri (approved):

Issue #480, opened 2022-01-11
Migration DB connection does not support special chars in password


ggarri:
### Describe the bug

QKM does not support DB password with special chars in it.

### Configuration

Set up a PG server with a password with special chars, such as `kpRVYL8F6EWM-7%-d/VBHLMl`

### Steps to reproduce

1. Set `DB_PASSWORD=kpRVYL8F6EWM-7%-d/VBHLMl`
2. Start QKM
3. Check migration status

**Actual result**

Migration fails to complete

**Expected result**

Migration completes successfully and QKM boot up correctly

**Logs**

```
time="2022-01-11T12:31:25Z" level=fatal msg="execution failed" error="parse \"postgres://qkm:kpRVYL8F6EWM-7%-d/VBHLMl@mainpsqlserver.postgres.database.azure.com:5432/qkm?sslcert=&sslkey=&sslmode=require&sslrootcert=\": invalid port \":kpRVYL8F6EWM-7%-d\" after host"
```

**Other**
https://consensys.slack.com/archives/C02T2TG4CKC/p1641904876005400

PR #489, opened 2022-01-18
support no url encoded db passwords

ggarri:
## Fixed Issue(s)
#480 

## Changelog

### DIFF
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -147,7 +147,14 @@ func migrateReset() error {
 
 func initMigrations(vipr *viper.Viper, logger log.Logger) (*migrate.Migrate, error) {
 	pgCfg := flags.NewPostgresConfig(vipr)
-	dbURL := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?", pgCfg.User, pgCfg.Password, pgCfg.Host, pgCfg.Port, pgCfg.Database)
+	var userPass string
+	if pgCfg.Password != "" {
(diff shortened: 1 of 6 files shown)


cmd/migrate.go
```diff
@@ -147,7 +147,14 @@ func migrateReset() error {
 
 func initMigrations(vipr *viper.Viper, logger log.Logger) (*migrate.Migrate, error) {
 	pgCfg := flags.NewPostgresConfig(vipr)
-	dbURL := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?", pgCfg.User, pgCfg.Password, pgCfg.Host, pgCfg.Port, pgCfg.Database)
+	var userPass string
+	if pgCfg.Password != "" {
```
ggarri, line 151: Password is not a mandatory value since you can configure PG to allow internal connections or other similar rules based on source

cmd/migrate.go
```diff
@@ -147,7 +147,14 @@ func migrateReset() error {
 
 func initMigrations(vipr *viper.Viper, logger log.Logger) (*migrate.Migrate, error) {
 	pgCfg := flags.NewPostgresConfig(vipr)
-	dbURL := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?", pgCfg.User, pgCfg.Password, pgCfg.Host, pgCfg.Port, pgCfg.Database)
+	var userPass string
```
6l20, line 150: I would name this one userCredentials

6l20 (approved):

cmd/migrate.go
ggarri: They are not always credentials since you might only have username in it

PR #491, opened 2022-01-21
enabling sonarcloud

ggarri:
## Changelog


sonar-project.properties
```diff
@@ -0,0 +1,19 @@
+sonar.projectKey=ConsenSys_quorum-key-manager
+sonar.organization=consensys
+sonar.projectVersion=21.12
+
+# relative paths to source directories. More details and properties are described
+# in https://sonarcloud.io/documentation/project-administration/narrowing-the-focus/
+sonar.sources=./src,./pkg
+sonar.exclusions=**/*_test.go,./tests/*,./cmd/*,**/mock/*,**/mocks/*,**/testutils/*,**/testdata/*,./src/infra/akv/*,./src/infra/aws/*
```
6l20, line 8: Why are these infra bits excluded ?

6l20 (approved):

sonar-project.properties
ggarri: I excluded aws and akv infra source code because there is not way we can test that in our integration test or unit tests

Issue #485, opened 2022-01-17
[AWS KMS] Key Creation Race condition


6l20:
@6l20 commented on [Wed Dec 01 2021](https://github.com/ConsenSys/orchestrate/issues/755)

### Describe the bug

Account creation fails with a dependency error ff2000 first time request is sent.
More generally, Key creation process fails when the key state is not appropriate and thus rejects the `GetPublicKey` call.

### Configuration

Spin up a fresh installation of orchestrate with qkm 21.9.3 with orchestrate helm chart 1.0.6

### Steps to reproduce

1. Instanciate QKM with  an AWS key store
2. Create a new account
```
curl --location --request POST 'localhost:8080/stores/aws-keys/keys/{demo-key-1}' \
--header 'Content-Type: application/json' \
--header 'X-API-Key: <API Key>' 
```
**Actual result**

Status : 500

Response body:

`{"message":"FF200@: Internal server error. Please ask an admin for help or try again later","code":1044992}`

**Expected result**

Status: 200

+ An account creation success response

### Additional context

**Logs**

`{"@timestamp":"2021-11-30T15:12:14Z","alias":"personal-account-rinkeby","error":"Error making API request.\nCode: . {\"message\":\"ST100: failed to get AWS key\",\"code\":\"ST100\"}:\nStatus: 404.","log.level":"error","message":"(use-cases.create-account): failed to import/create ethereum account"}`

**Notes**

**Works when request is sent twice**

**Works with an AKV store**

---

@6l20 commented on [Wed Jan 12 2022](https://github.com/ConsenSys/orchestrate/issues/755#issuecomment-1010773387)

This problem appears to be linked to some race condition in the KMS backend.

According to the Keys state
https://docs.aws.amazon.com/kms/latest/developerguide/key-state.html

We have a `KMSInvalidStateException` when calling the `GetPublicKey`API right after `CreateKey`.

Suggested solution is to use the `DescribeKey`API and "wait" the Enabled flag to come up, inbetween above calls.

https://docs.aws.amazon.com/kms/latest/APIReference/API_DescribeKey.html

---

@6l20 commented on [Mon Jan 17 2022](https://github.com/ConsenSys/orchestrate/issues/755#issuecomment-1014631068)

Here below is the answer made by AWS support

Hello Sylvain,

Hope this email finds you well. I understand you have found matching CloudTrail logs for the errors you've noticed.

[...]

PR #493, opened 2022-01-21
485 aws retry

ggarri:
## Fixed Issue(s)
#485 

## Changelog

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.16 // indirect
 	github.com/auth0/go-jwt-middleware/v2 v2.0.0-beta.1
 	github.com/aws/aws-sdk-go v1.38.41
+	github.com/cenkalti/backoff/v4 v4.0.2
 	github.com/consensys/gnark-crypto v0.5.0
 	github.com/consensys/quorum v2.7.0+incompatible
 	github.com/containerd/containerd v1.4.11 // indirect
--- a/src/infra/aws/client/kms.go
+++ b/src/infra/aws/client/kms.go
@@ -18,6 +18,8 @@ const (
 	KeyStatePendingDeletion = "PendingDeletion"
 	KeyStatePendingImport   = "PendingImport"
 	KeyStateUpdating        = "Updating"
+	KeyStateUnavailable     = "Unavailable"
+	KeyStateDisabled        = "Disabled"
 )
 
 func (c *AWSClient) CreateKey(ctx context.Context, keyID, keyType string, tags []*kms.Tag) (*kms.CreateKeyOutput, error) {
@@ -41,7 +43,12 @@ func (c *AWSClient) CreateKey(ctx context.Context, keyID, keyType string, tags [
 		return nil, parseKmsErrorResponse(err)
 	}
 
-	err = c.waitDuringState(ctx, keyID, KeyStateCreating)
+	err = c.waitKeyState(ctx, keyID, func(metadata *kms.KeyMetadata) error {
+		if *metadata.Enabled {
+			return nil
+		}
+		return fmt.Errorf("key %s is still in not enabled", keyID)
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -141,7 +148,13 @@ func (c *AWSClient) DeleteKey(ctx context.Context, keyID string) (*kms.ScheduleK
 		return nil, parseKmsErrorResponse(err)
 	}
 
-	err = c.waitDuringState(ctx, keyID, KeyStatePendingDeletion)
+	err = c.waitKeyState(ctx, keyID, func(metadata *kms.KeyMetadata) error {
+		if !*metadata.Enabled && *metadata.KeyState == KeyStatePendingDeletion {
+			return nil
+		}
+
+		return fmt.Errorf("key %s was not scheduled for deletion yet", keyID)
+	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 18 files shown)


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.16 // indirect
 	github.com/auth0/go-jwt-middleware/v2 v2.0.0-beta
 	github.com/aws/aws-sdk-go v1.38.41
+	github.com/cenkalti/backoff/v4 v4.0.2
```
6l20, line 15: Do we really need exponential backoff to solve that issue and/or an additional dependency ?

src/infra/aws/client/kms.go
```diff
@@ -116,25 +140,63 @@ func (c *AWSClient) Sign(_ context.Context, keyID string, msg []byte, signingAlg
...
+		if !*metadata.Enabled && *metadata.KeyState == KeyStatePendingDeletion {
+			return nil
+		}
+
+		return fmt.Errorf("key %s was not scheduled for deletion yet", keyID)
+	})
+	if err != nil {
+		return nil, err
+	}
+
```
6l20, line 161: Here according to the state table, the operation will fail if the state is KeyPendingDeletion or Updating.
And in the case KeyPendingDeletion that should mean that deletion has already been scheduled

src/infra/aws/client/kms.go
```diff
@@ -116,25 +140,63 @@ func (c *AWSClient) Sign(_ context.Context, keyID string, msg []byte, signingAlg
...
 		KeyId: &keyID,
 	})
 	if err != nil {
 		return nil, parseKmsErrorResponse(err)
 	}
 
+	err = c.waitKeyState(ctx, keyID, func(metadata *kms.KeyMetadata) error {
+		if *metadata.KeyState == KeyStatePendingDeletion {
+			return fmt.Errorf("key %s is still pending for deletion deletion", keyID)
+		}
```
6l20, line 176: Here it will only be accepted in that case, so we should only accept that state

go.mod
ggarri: I think using a retry backoff strategy is the best way to go considering the indeterministic behaviour in terms of time of AWS. THis library is considering fairly popular and light so I don't think it worths the effort of rewriting the same logic myself.

src/infra/aws/client/kms.go
ggarri: the operation will fail in meanwhile the key is still enabled and not scheduled for deletion. I am not sure I understand what you are proposing, would u mind to rephrase the concern?

src/infra/aws/client/kms.go
ggarri: yes, what else should I do instead?

src/infra/aws/client/kms.go
6l20: Ok I know where my mistake is, I missed the fact that you were calling that check after invoking the ScheduleForDeletion, so here is ok. 
But should not it be called before ? 
The errors we have are on that call being rejected, because key has already been scheduled for deletion

src/infra/aws/client/kms.go
6l20: This comment was supposed to highlight the misunderstanding above, my mistake. Thats ok

src/vaults/service/vaults/create_aws_test.go
```diff
@@ -0,0 +1,35 @@
...
+	defer ctrl.Finish()
+
+	logger := testutils.NewMockLogger(ctrl)
+	roles := mock.NewMockRoles(ctrl)
+	vault := New(roles, logger)
+
+	ctx := context.Background()
+	vaultName := "aws-vault"
+	cfg := &entities.AWSConfig{}
+	allowedTenants := []string{"tenant_id_1"}
```
6l20, line 26: "tenant_id_1" should be a const here

src/vaults/service/vaults/create_hashicorp_test.go
```diff
@@ -0,0 +1,35 @@
...
+	defer ctrl.Finish()
+
+	logger := testutils.NewMockLogger(ctrl)
+	roles := mock.NewMockRoles(ctrl)
+	vault := New(roles, logger)
+
+	ctx := context.Background()
+	vaultName := "hashicorp-vault"
+	cfg := &entities.HashicorpConfig{}
+	allowedTenants := []string{"tenant_id_1"}
```
6l20, line 26: "tenant_id_1" should be a const

src/vaults/service/vaults/get_test.go
```diff
@@ -0,0 +1,64 @@
...
+	vault := New(roles, logger)
+
+	ctx := context.Background()
+	vaultName := "vault-id"
+	cfg := &entities.AWSConfig{}
+	allowedTenants := []string{"tenant_id_1"}
+
+	t.Run("should create and get vault client successfully", func(t *testing.T) {
+		userInfo := &entities2.UserInfo{
+			Tenant: "tenant_id_1",
```
6l20, line 31: "tenant_id_1" and "tenant_id_2" should be consts

src/vaults/service/vaults/create_aws_test.go
ggarri: done

src/vaults/service/vaults/get_test.go
ggarri: done

src/vaults/service/vaults/create_hashicorp_test.go
ggarri: done

Issue #500, opened 2022-02-10
Add support of Curve25519 in the keys


ggarri:
### Summary / User Story

Add support of Curve25519 in the keys abstraction of the QKM in order to support WavesNetwork
  
### Technical Notes

https://github.com/golang/crypto/blob/master/curve25519/curve25519.go

## Business Value
Extend the amount of integration with new potential users such as WavesNetwork

### Acceptance Criteria

- [x] Users can create keys using curve25519
- [x] Users can import keys using curve25519
- [x] Users can sign data using curve25519
- [x]  Users can verify data using curve25519

ggarri:
Documentation regarding supported key types on `WavesNetwork`
https://docs.waves.tech/en/blockchain/waves-protocol/cryptographic-practical-details#creating-a-private-key-from-a-seed

PR #504, opened 2022-02-11
implement support for ed25519

ggarri:
## Fixed Issue(s)
#500 

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,11 @@
 ## v21.12.2 (Unreleased)
 ### 🆕 Features
 * Support for OIDC token custom claims `AUTH_OIDC_CUSTOM_CLAIMS` for tenant_id and permissions.
-  
+* Support for ED25519 keys (EdDSA signature scheme using SHA-512 and Curve25519)
+
 ### 🛠 Bug fixes
 * Invalid authentication forwarded to downstream proxy nodes if QKM authentication is enabled.
 * Fixed AWS client issue preventing racing condition on key state transition changes. 
-* Fixed invalid request error on create Registry with empty payload
-* Fixed forwarding of vault service 429 errors.
 
 ## v21.12.1 (2021-12-20)
 ### 🛠 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@
 
 ### 🛠 Bug fixes
 * Invalid authentication forwarded to downstream proxy nodes if QKM authentication is enabled.
-* Fixed AWS client issue preventing racing condition on key state transition changes. 
+* Fixed AWS client issue preventing racing condition on key state transition changes.
+* Fixed invalid request error on create Registry with empty payload
+* Fixed forwarding of vault service 429 errors.
 
 ## v21.12.1 (2021-12-20)
 ### 🛠 Bug fixes
--- a/pkg/crypto/ecdsa/secp256k1.go
+++ b/pkg/crypto/ecdsa/secp256k1.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/consensys/quorum-key-manager/pkg/errors"
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
@@ -30,17 +29,17 @@ func CreateSecp256k1(importedPrivKey []byte) (privKey, pubKey []byte, err error)
 
 func SignSecp256k1(privKey, data []byte) ([]byte, error) {
 	if len(data) != crypto.DigestLength {
-		return nil, errors.InvalidParameterError(fmt.Sprintf("data is required to be exactly %d bytes (%d)", crypto.DigestLength, len(data)))
+		return nil, fmt.Errorf("data is required to be exactly %d bytes (%d)", crypto.DigestLength, len(data))
 	}
 
 	ecdsaPrivKey, err := crypto.ToECDSA(privKey)
 	if err != nil {
-		return nil, errors.InvalidParameterError(fmt.Sprintf("failed to parse private key. %s", err.Error()))
+		return nil, fmt.Errorf("failed to parse private key. %s", err.Error())
 	}
 
 	signature, err := crypto.Sign(data, ecdsaPrivKey)
 	if err != nil {
-		return nil, errors.CryptoOperationError(fmt.Sprintf("failed to sign. %s", err.Error()))
+		return nil, fmt.Errorf("failed to sign. %s", err.Error())
 	}
 
 	// We remove the recID from the signature (last byte).
@@ -53,7 +52,7 @@ func VerifySecp256k1Signature(publicKey, message, signature []byte) (bool, error
 		return false, err
 	}
 	if len(signature) != 64 {
-		return false, errors.InvalidParameterError("invalid secp256k1 signature length")
+		return false, fmt.Errorf("invalid secp256k1 signature length")
 	}
 
 	r := new(big.Int).SetBytes(signature[0:32])
--- a/pkg/crypto/ecdsa/secp256k1.go
+++ b/pkg/crypto/ecdsa/secp256k1.go
@@ -52,6 +52,9 @@ func VerifySecp256k1Signature(publicKey, message, signature []byte) (bool, error
 	if err != nil {
 		return false, err
 	}
+	if len(signature) != 64 {
+		return false, errors.InvalidParameterError("invalid secp256k1 signature length")
+	}
 
 	r := new(big.Int).SetBytes(signature[0:32])
 	s := new(big.Int).SetBytes(signature[32:64])
--- a/pkg/crypto/ecdsa/secp256k1.go
+++ b/pkg/crypto/ecdsa/secp256k1.go
@@ -1,4 +1,4 @@
-package crypto
+package ecdsa
 
 import (
 	"crypto/ecdsa"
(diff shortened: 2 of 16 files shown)


src/stores/store/keys/hashicorp/hashicorp.go
```diff
@@ -41,6 +41,12 @@ func New(client hashicorp.PluginClient, logger log.Logger) *Store {
 }
 
 func (s *Store) Create(_ context.Context, id string, alg *entities2.Algorithm, attr *entities.Attributes) (*entities.Key, error) {
+	if !s.isSupportedAlgo(alg) {
+		errMessage := "invalid or not supported elliptic curve and signing algorithm for Hashicorp key creation"
+		s.logger.With("elliptic_curve", alg.EllipticCurve, "signing_algorithm", alg.Type).Error(errMessage)
+		return nil, errors.NotSupportedError(errMessage)
```
darioAnongba, line 47: Are we sure we want to return `NotSupported` ? I think there is a difference between these 3 error types:
- Not Implemented means that the pair curve/algo exists but it's not implemented for this store
- Not Supported means that the pair curve/algo exists but this store does not support it
- Invalid Parameter means the pair curve/algo does not exist (invalid).

Here it seems we want InvalidParameter

src/stores/store/keys/hashicorp/hashicorp.go
```diff
@@ -52,11 +58,16 @@ func (s *Store) Create(_ context.Context, id string, alg *entities2.Algorithm, a
...
 	}
-
 	return parseAPISecretToKey(res)
 }
 
 func (s *Store) Import(_ context.Context, id string, privKey []byte, alg *entities2.Algorithm, attr *entities.Attributes) (*entities.Key, error) {
+	if !s.isSupportedAlgo(alg) {
+		errMessage := "invalid or not supported elliptic curve and signing algorithm for Hashicorp key import"
+		s.logger.With("elliptic_curve", alg.EllipticCurve, "signing_algorithm", alg.Type).Error(errMessage)
+		return nil, errors.NotSupportedError(errMessage)
```
darioAnongba, line 68: Are we sure we want to return `NotSupported` ? I think there is a difference between these 3 error types:
- Not Implemented means that the pair curve/algo exists but it's not implemented for this store
- Not Supported means that the pair curve/algo exists but this store does not support it
- Invalid Parameter means the pair curve/algo does not exist (invalid).

Here it seems we want InvalidParameter

src/stores/store/keys/hashicorp/hashicorp.go
```diff
@@ -165,7 +176,13 @@ func (s *Store) Destroy(_ context.Context, id string) error {
 	return nil
 }
 
-func (s *Store) Sign(_ context.Context, id string, data []byte, _ *entities2.Algorithm) ([]byte, error) {
+func (s *Store) Sign(_ context.Context, id string, data []byte, alg *entities2.Algorithm) ([]byte, error) {
+	if !s.isSupportedAlgo(alg) {
+		errMessage := "invalid or not supported elliptic curve and signing algorithm for Hashicorp signing"
+		s.logger.With("elliptic_curve", alg.EllipticCurve, "signing_algorithm", alg.Type).Error(errMessage)
+		return nil, errors.NotSupportedError(errMessage)
```
darioAnongba, line 183: Are we sure we want to return `NotSupported` ? I think there is a difference between these 3 error types:
- Not Implemented means that the pair curve/algo exists but it's not implemented for this store
- Not Supported means that the pair curve/algo exists but this store does not support it
- Invalid Parameter means the pair curve/algo does not exist (invalid).

Here it seems we want InvalidParameter

src/stores/connectors/keys/keys.go
```diff
@@ -35,5 +35,9 @@ func isSupportedAlgo(alg *entities.Algorithm) bool {
 		return true
 	}
 
+	if alg.Type == entities.Eddsa && alg.EllipticCurve == entities.X25519 {
```
darioAnongba, line 38: how come this function exists? Shouldn't we have specific functions for each store?

src/stores/store/keys/hashicorp/hashicorp.go
```diff
@@ -192,3 +209,15 @@ func (s *Store) Encrypt(ctx context.Context, id string, data []byte) ([]byte, er
 func (s *Store) Decrypt(ctx context.Context, id string, data []byte) ([]byte, error) {
 	return nil, errors.ErrNotImplemented
 }
+
+func (s *Store) isSupportedAlgo(alg *entities2.Algorithm) bool {
```
darioAnongba, line 213: Missing implementation in the plugin?

pkg/crypto/eddsa/x25519.go
```diff
@@ -0,0 +1,53 @@
+package eddsa
```
darioAnongba, line 1: this package should not exist. As you can see these are utility functions of the store and should be used only once in the codebase. A code smell is that you are returning specific error types from this package.

pkg/crypto/eddsa/babyjubjub.go
```diff
@@ -0,0 +1,69 @@
+package eddsa
```
darioAnongba, line 1: this package should not exist. As you can see these are utility functions of the store and should be used only once in the codebase. A code smell is that you are returning specific error types from this package.

pkg/crypto/ecdsa/secp256k1.go
```diff
@@ -0,0 +1,63 @@
+package ecdsa
```
darioAnongba, line 1: this package should not exist. As you can see these are utility functions of the store and should be used only once in the codebase. A code smell is that you are returning specific error types from this package.

src/stores/store/keys/local/local.go
```diff
@@ -218,16 +214,26 @@ func (s *Store) Sign(ctx context.Context, id string, data []byte, algo *entities
...
+		signature, err = ecdsa.SignSecp256k1(privkey, data)
+	case algo.Type == entities2.Eddsa && algo.EllipticCurve == entities2.X25519:
+		signature, err = eddsa.SignX25519(privkey, data)
 	default:
 		errMessage := "signing algorithm and curve combination not supported for signing"
 		logger.With("algorithm", algo.Type, "curve", algo.EllipticCurve).Error(errMessage)
 		return nil, errors.InvalidParameterError(errMessage)
 	}
+
+	if err != nil {
```
darioAnongba, line 231: I think doing `return` directly in the switches was correct because these functions should not come from a package and return a specific error type. If you want to move them to a package, they should return a standard error and this `if err != nil` should return the specific error (but you won't know what happened)

CHANGELOG.md
```diff
@@ -3,12 +3,11 @@
...
 ### 🆕 Features
 * Support for OIDC token custom claims `AUTH_OIDC_CUSTOM_CLAIMS` for tenant_id and permissions.
-  
+* Support for ED25519 keys (EdDSA signature scheme using SHA-512 and Curve25519)
+
 ### 🛠 Bug fixes
 * Invalid authentication forwarded to downstream proxy nodes if QKM authentication is enabled.
 * Fixed AWS client issue preventing racing condition on key state transition changes. 
-* Fixed invalid request error on create Registry with empty payload
-* Fixed forwarding of vault service 429 errors.
```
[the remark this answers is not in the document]
ggarri: merging issues, undoing

src/entities/algo.go
```diff
@@ -9,6 +9,7 @@ const (
 
 	Babyjubjub Curve = "babyjubjub"
 	Secp256k1  Curve = "secp256k1"
+	X25519     Curve = "x25519"
```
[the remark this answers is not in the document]
ggarri: Since everything is a curve I didn't want to prefix it by that and checking around I see many references to `x25519` to the curve itself https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x25519/ .

Even the documentation of the client who request it also refers to it that way
https://docs.waves.tech/en/blockchain/waves-protocol/cryptographic-practical-details#creating-a-private-key-from-a-seed

src/stores/store/keys/hashicorp/hashicorp.go
```diff
@@ -41,6 +41,12 @@ func New(client hashicorp.PluginClient, logger log.Logger) *Store {
 }
 
 func (s *Store) Create(_ context.Context, id string, alg *entities2.Algorithm, attr *entities.Attributes) (*entities.Key, error) {
+	if !s.isSupportedAlgo(alg) {
```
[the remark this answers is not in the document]
ggarri: Yes, I was also surprised we did not caught this before

src/stores/connectors/keys/create_test.go
```diff
@@ -24,6 +24,10 @@ func TestCreateKey(t *testing.T) {
 	defer ctrl.Finish()
 
 	key := testutils2.FakeKey()
+	key.Algo = &entities2.Algorithm{
```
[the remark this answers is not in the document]
ggarri: I actually added more tests, you can see the coverage is 91%, this change just enforces a specific key type which it is not part of the previous tests

pkg/crypto/eddsa/babyjubjub.go
ggarri: closing in favor of previous comment

ggarri: closing in favor of previous comment

pkg/crypto/ecdsa/secp256k1.go
ggarri: closing in favor of previous comment

darioAnongba (approved):

src/stores/store/keys/local/local.go
ggarri: Discussed

pkg/crypto/eddsa/x25519.go
ggarri: discussed

PR #541, opened 2022-03-02
Fix padding occuring when applying malleability fix

darioAnongba:
Fix padding occuring when applying malleability fix

### DIFF
--- a/src/infra/log/zap/logger.go
+++ b/src/infra/log/zap/logger.go
@@ -87,21 +87,7 @@ func (l Logger) WithComponent(component string) log.Logger {
 }
 
 func (l *Logger) Write(p []byte) (n int, err error) {
-	switch l.cfg.Level {
-	case DebugLevel:
-		l.Debug(string(p))
-	case InfoLevel:
-		l.Info(string(p))
-	case WarnLevel:
-		l.Warn(string(p))
-	case ErrorLevel:
-		l.Error(string(p))
-	case PanicLevel:
-		l.Panic(string(p))
-	default:
-		l.Info(string(p))
-	}
-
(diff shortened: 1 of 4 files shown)


src/infra/log/zap/logger.go
```diff
@@ -87,21 +87,7 @@ func (l Logger) WithComponent(component string) log.Logger {
 }
 
 func (l *Logger) Write(p []byte) (n int, err error) {
-	switch l.cfg.Level {
```
ggarri, line 90: Did you validate this change? IIRC this implementation was intended to only Write logs if level corresponds to the expected one. Otherwise you will write logs even if the defined logger level does not match

tests/e2e/ethereum_test.go
```diff
@@ -278,6 +281,16 @@ func (s *ethTestSuite) TestSignTransaction() {
 		httpError := err.(*client.ResponseError)
 		assert.Equal(s.T(), 404, httpError.StatusCode)
 	})
+
+	s.Run("should sign a big amount of transactions successfully", func() {
+		for i := 0; i < 500; i++ {
```
ggarri, line 286: What is the added value of this test unit? If we are confident about the fix we did there is not need for it. WDYT? 

In addition I would rather evaluate the long term cost of this before pushing

darioAnongba: removed

src/infra/log/zap/logger.go
darioAnongba: yes I validated this, it was a bug before that printed the access logs at the same level as the config.

ggarri (approved):

ggarri (approved):

